PR #4, opened 2019-06-25
Make setup works like data

antfu:
In the RFC https://github.com/vuejs/rfcs/blob/function-apis/active-rfcs/0000-function-api.md#state

> This works exactly like data() - msg becomes a reactive and mutable property

**Changes**
- Plain values returned  in `setup()` become reactive automatically
- Unit test added


src/setup.ts
```diff
@@ -109,4 +111,23 @@ export function mixin(Vue: VueConstructor) {
     }
     return ctx;
   }
+
+  function makeBindingReactive(binding: any) {
```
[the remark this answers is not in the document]
antfu: But in this case the `state` and `value` will be called outside of `setup()`

https://github.com/vuejs/vue-function-api/blob/109b877a0f6e87cd6e508e36b30831ef382faece/src/setup.ts#L41-L43

antfu:
Fixed

src/setup.ts
```diff
@@ -58,7 +57,17 @@ export function mixin(Vue: VueConstructor) {
...
 
     Object.keys(binding).forEach(name => {
-      const bindingValue = binding[name];
+      let bindingValue = binding[name];
+      if (bindingValue === undefined)
+        return;
+      // make plain value reactive
+      if (!isWrapper(bindingValue) && !isFunction(bindingValue) && !Object.isFrozen(bindingValue)) {
+        if (isPlainObject(bindingValue))
+          bindingValue = state(bindingValue);
```
liximomo: We should always use `value`.  Consider the example below. The assignment to `form` in the template will not trigger a re-render if we use `state`.

```js
{
  setup() {
    return {
      form: { a:'', b: ''}
    }
  },
   template: `
      <div>
        a: {{ form.a }} b: {{ form.b }}
        <button @click="form = { a: 1, b: 2}"><button>
      <div>
    `
}
```

antfu: I have updated the test in 950caa1ba01278d21b654a9b49eb3a707a91787f
```js
  it('should make returned plain value reactive', done => {
    const vm = new Vue({
      setup() {
        return {
          name: null,
          nested: {
            object: {
              msg: 'foo'
            }
          },
          foo: {
            bar: 0
          }
        };
      },
      template: '<div>{{ name }}, {{ nested.object.msg }}, {{ foo.bar }}</div>',
    }).$mount();
    expect(vm.$el.textContent).toBe(', foo, 0');
    vm.name = 'foo';
    vm.nested.object.msg = 'bar';
    vm.foo = { bar: 1 };
    waitForUpdate(() => {
      expect(vm.$el.textContent).toBe('foo, bar, 1');
    }).then(done);
  })
```
Both `state` and `value` approaches seems passed the test
```ts
      // make plain value reactive
      if (!isWrapper(bindingValue) && !isFunction(bindingValue) && !Object.isFrozen(bindingValue)) {
        if (isPlainObject(bindingValue))
          bindingValue = state(bindingValue);
        else
          bindingValue = value(bindingValue);
      }
```
```ts
      // make plain value reactive
      if (!isWrapper(bindingValue) && !isFunction(bindingValue) && !Object.isFrozen(bindingValue)) {
        bindingValue = value(bindingValue);
      }
```
I am not sure whether these two approaches have difference in performance on re-rendering or not. 
I would use `value` alone for now.

liximomo:
```js
// these can trigger the re-render
vm.name = 'foo';
vm.nested.object.msg = 'bar';

// this can't
vm.foo = { bar: 1 };
```

Change the test to: 
```js
  it('should make returned plain value reactive', done => {
    const vm = new Vue({
      setup() {
        return {
          foo: {
            bar: 0
          }
        };
      },
      template: '<div>{{ foo.bar }}</div>',
    }).$mount();
    expect(vm.$el.textContent).toBe('0');
    vm.foo = { bar: 1 };
    waitForUpdate(() => {
      expect(vm.$el.textContent).toBe('1');
    }).then(done);
  })
```

antfu: @liximomo Thanks for point out. You are right. I will update the test soon.

antfu: @liximomo Just updated the test. 

Now we can address the problem of `state`. Using `value` works fine. 

Thanks for the detailed reply.

PR #12, opened 2019-06-28
Requires minimum vue version 2.6.x

antfu:
This project is using [`observable` API](https://vuejs.org/v2/api/#Vue-observable), which is introduced in vue 2.6.0+

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,6 +29,8 @@ Future-Oriented Programming, `vue-function-api` provides function api from `Vue3
 
 # Installation
 
+> `vue-function-api` requires minimum version `2.6.x` of `vue` as peer dependency
+
 **npm**
 ```bash
 npm install vue-function-api --save
--- a/README.zh-CN.md
+++ b/README.zh-CN.md
@@ -29,6 +29,8 @@
 
 # 安装
 
+> `vue-function-api` 需要同时装有 `2.6.x` 版本以上的 `vue`
+
 **npm**
 ```bash
 npm install vue-function-api --save
--- a/package.json
+++ b/package.json
@@ -40,7 +40,7 @@
     "typescript": "^3.4.5"
   },
   "peerDependencies": {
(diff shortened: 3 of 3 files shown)


liximomo:
`Vue.observable` is not a must-have thing.

```js
export function observable<T = any>(obj: T): T {
  const Vue = getCurrentVue();
  if (Vue.observable) {
    return Vue.observable(obj);
  }

  const silent = Vue.config.silent;
  Vue.config.silent = true;
  const vm = new Vue({
    data: {
      $$state: obj,
    },
  });
  Vue.config.silent = silent;

  return vm._data.$$state;
}

```

antfu:
As a future-oriented library. Should we encourage users to use the latest version of vue2? 

There are some other APIs used in this library were introduced after 2.0:

- [`inject`/`provide`](https://vuejs.org/v2/api/#provide-inject) is introduced in 2.2+
- [`vm.$props`](https://vuejs.org/v2/api/#vm-props) and `vm.$attrs` is introduced in 2.4+
- [`observable`](https://vuejs.org/v2/api/#Vue-observable) is introduced in 2.6+

I have tried installing different version of vue on my local machine and running the tests. Versions from 2.0.x~2.4.x will somehow failed the tests. The minimum working version is 2.5.x.

Also backward compatibility will make code more complex and having more issues to handle. Which to me is kinda unnecessary in the library.

antfu:
We could use `Vue.version` to check and give user a runtime warning of outdated vue.

liximomo:
> We could use Vue.version to check and give user a runtime warning of outdated vue.
That's a good idea.

liximomo:
The required minimum version of vue has changed to v2.5.22.

antfu:
@liximomo Hi I just saw the latest commit, I would suggest to update version of vue in peerDependencies. The version of devDependencies would not affect on user side.

liximomo:
Thanks!

PR #23, opened 2019-07-03
Add onBeforeCreate Lifecycle

tochoromero:
It seems we are missing the `onBeforeCreate` lifecycle hook


tochoromero:
So... it seems the `beforeCreate` hook can not be registered with `vm.$on('hook:beforeCreate', callback);`
I guess that is why it was not included in the first place.

Is this a limitation of the 2.x port? I would assume the 3.x version will include the hook.

liximomo:
You can think `setup` itself as the `beforeCreate` lifecycle.

PR #37, opened 2019-07-24
feat: add an option to bind other props

kidonng:
This PR makes the following possible:

```js
import Vue from 'vue'
import VueRouter from 'vue-router'
import { plugin } from 'vue-function-api'

Vue.use(plugin, { props: ['router'] })

const router = new VueRouter()

new Vue({
  router,
  setup(props, context) {
    context.router === context.root.$router // true
  }
})
```

I'm not familiar with TypeScript, feel free to correct me :P

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,7 +57,7 @@ import { plugin } from 'vue-function-api'
 
 Vue.use(plugin)
 
-// Optionally pass in some options
+// Optionally, pass in some options
 Vue.use(plugin, { props: ['router'] })
 ```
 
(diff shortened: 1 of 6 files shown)


posva:
I would avoid adding any new feature that is not covered by the RFC as the point of this lib is to allow users play with the api before it is available and make it very easy to migrate to the function api once it land on Vue 3. Adding features that will not be added to Vue 3 would go against this

kidonng:
Sorry, I was reading the translated SFC which doesn't cover the `context` parameter and I thought it was provided by `vue-function-api` as a helper because we can't access `this` in `setup()`.

posva:
No, no need to be sorry! I know it came from a good intention 😀
But it's important to keep the RFC into consideration when directly creating PR for new features

PR #42, opened 2019-07-30
Expose isWrapper in export

AlbertBrand:
For AlbertBrand/vue-async-function#1 I want to check whether a value is wrapped. There is already a function for it but it's not exposed. This PR exposes the `isWrapper` function so I can start using it.


posva:
Same as https://github.com/vuejs/vue-function-api/pull/37#issuecomment-516844717. This should be considered as part of the RFC first

PR #46, opened 2019-08-02
fix: fix recursive fire initSetup in safeRunCb

iwyvi:
I found when assign a `value` in `watch` with `lazy = false`, it will fire the callback (`initSetup`) in `safeRunCb` recursively.

Reproduce code: 

```javascript
new Vue({
  setup() {
    const a = value(1);
    const b = value(9);
    watch(a, () => {
      b.value++;
    });
    return { a, b };
  },
  template: `<div>{{a}} {{b}}</div>`
}).$mount();
```

I think the reason is when touches `b` in watch's callback, the `b` will be collected by the scope of `dep.Target` created in `safeRunCb`, then its value changes and will trigger an update in that scope. And it becomes an infinite loop.

### DIFF
--- a/src/setup.ts
+++ b/src/setup.ts
@@ -12,11 +12,15 @@ let disableSetup = false;
 // `cb` should be called right after props get resolved
 function waitPropsResolved(vm: VueInstance, cb: (v: VueInstance, props: Record<any, any>) => void) {
   const safeRunCb = (props: Record<any, any>) => {
+    let callOnce = false;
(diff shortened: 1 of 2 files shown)


kidonng:
Confirmed. Actually you don't even need to assign value in `watch()` to create a loop:

```js
const foo = value('bar')
watch(() => foo, v => {
  console.log(v)
})
foo.value = 'baz'
```

liximomo:
How `setup` is invoked is going to change. I think this won't happen after the change.

PR #48, opened 2019-08-05
Add test verifying mixins don't break computed properties

thecrypticace:
In the current version of vue-function-api (2.0.6) if you define a mixin with computed properties usages of `computed()` will break when you have more than one component of a given type (like in a v-for list). The computed properties will return the values from the first component even for all other instances.

The master branch does **not** have this problem (awesome! — do you know when a release will be tagged?) but I thought having an explicit test for this would be useful to ensure no regressions occur.


liximomo:
Confirmed.

PR #58, opened 2019-08-12
Fix storage of DOM elements

thecrypticace:
If you store a DOM element in a `value()` you may get errors from the browser or in more rare cases (explained in the commit message) an error during vnode patching.

The vnode patching error feels like something inside Vue is wrong for something like this to be able to cause that kind of error but I want to look more into it to see why that might be happening.

In any case this appears to fix the issue from what I can tell. I have not written a test for it yet, just a stub. I need to figure out how to write one that will verify the problem whether or not its run in a browser.

### DIFF
--- a/src/reactivity/observable.ts
+++ b/src/reactivity/observable.ts
@@ -1,6 +1,6 @@
 import { AnyObject } from '../types/basic';
 import { getCurrentVue } from '../runtimeContext';
-import { isObject, def, hasOwn } from '../utils';
+import { isObject, def, hasOwn, isDOM } from '../utils';
 import { isWrapper } from '../wrappers';
 import { AccessControIdentifierlKey, ObservableIdentifierKey } from '../symbols';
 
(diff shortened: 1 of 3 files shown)


beeplin:
Cool! I met this problem yesterday when using tiptap rich text editor.

thecrypticace:
Oh good so I'm not completely crazy for trying to put a DOM element into an observed property.

🙌

liximomo:
The default environment in Jest is `jsdom`, we don't need to install it again.

beeplin:
Hi @liximomo, thank you for your efforts for the wonderful 2.2.0~  I am eager to try the new render function support ;) .

However,  just like what @thecrypticace shows above, when I am using tiptap (https://github.com/scrumpy/tiptap, probably the best rich text editor in
 Vue ecosystem) it breaks the whole vue runtime when emitting events under higher than 2.0.6, saying no el in vnode.

thecrypticace:
Seeing as the new proposal has changed things a bit I'm closing this PR. This issue will need to be resolved in some manner in any case due to the introduction of template refs.

PR #73, opened 2019-08-23
fix typos and removing couple typescript `as any`

pikax:
Typos and remove unnecessary `as any`

### DIFF
--- a/src/setup.ts
+++ b/src/setup.ts
@@ -51,16 +51,16 @@ function updateTemplateRef(vm: ComponentInstance) {
   }
 
   const newKeys = Object.keys(refs);
-  const validNewKyes = [];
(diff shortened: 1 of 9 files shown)


liximomo:
@pikax Can you move the "removing `as any`" things to a separate PR?

pikax:
done

src/setup.ts
```diff
@@ -146,16 +146,18 @@ export function mixin(Vue: VueConstructor) {
     if (!binding) return;
 
     if (isFunction(binding)) {
+      // keep typescript happy with the binding type.
```
liximomo: You missed here.

pikax: fixed it, thanks :)

PR #76, opened 2019-08-23
remove some typescript `as any`

pikax:
Removing some `as any`


src/reactivity/set.ts
```diff
@@ -41,7 +41,7 @@ export function set<T>(target: any, key: any, val: T): T {
     target[key] = val;
     return val;
   }
-  const ob = (target as any).__ob__;
+  const ob = target.__ob__;
```
pikax: target is already `any`

PR #77, opened 2019-08-23
improve typing on createComponentInstance function

pikax:
`createComponentInstance` returns a Vue instance that was passed in the first argument, this change allows to get the typing right when we pass the options.

I'm using the same types in `vue/types/options`.

NOTE: I'm not entirely sure about this


pikax:
Outdated

PR #80, opened 2019-08-24
[feature]: add server prefetch hook

skyline0705:
Add `serverPrefetch` hook for vue-function-api.
In my project, we use this hook for SSR. I have tried it and looks work fine~~

### DIFF
--- a/src/apis/lifecycle.ts
+++ b/src/apis/lifecycle.ts
@@ -36,5 +36,6 @@ export const onBeforeUnmount = createLifeCycle('beforeDestroy');
 // only one event will be fired between destroyed and deactivated when an unmount occurs
 export const onUnmounted = createLifeCycles(['destroyed', 'deactivated'], genName('unmounted'));
 export const onErrorCaptured = createLifeCycle('errorCaptured');
+export const onServerPrefetch = createLifeCycle('serverPrefetch');
 export const onActivated = createLifeCycle('activated');
 export const onDeactivated = createLifeCycle('deactivated');


pi0:
Any updates on this? 

/cc @LinusBorg @posva @liximomo @TooBug

posva:
I imagine the PR would need a test.
Since SSR is yet to be added to https://vue-composition-api-rfc.netlify.com/api.html, I'm unsure about. To me it makes sense for this library to implement it as it's for 2.x anyway and also list it on the Readme

rebeccarich:
Any update on this one @posva @LinusBorg @liximomo @TooBug? :)
Would be great to get an idea of the roadmap for SSR in this plugin is before we commit to using it in our app. SSR is a requirement for us unfortunately

michael-zucchetta:
This would be really helpful for us too and I think it should be added

pi0:
I think we should also ensure no context sharing is happening during concurrent SSR requests.  [runtimeContext](https://github.com/vuejs/composition-api/blob/master/src/runtimeContext.ts#L20) depends on a global reference!

filrak:
Any updates on that? I would happily commit my time to push it forward if needed. This feature is a must-have for our project unfortunately ;/

Akryum:
@pi0 I'm not sure this would be a problem, the global variables are used only during `setup` as it's processing the different components. So for a different request it will be overridden by each component one at a time.

Akryum:
Thanks for the PR @skyline0705 ! Closing in favor of #198

PR #85, opened 2019-08-26
fix: type-inference error when ref() accept an array type or ref-like type.

wjw99830:
```typescript
const a = ref<any[]>([]);
```
Here type of `a` is `UnwrapRef<any>[]`, it results a type-error although it works fine during runtime.
Why ?
```typescript
type UnwrapRef<T> = T extends Ref<infer V>
  ? UnwrapRef2<V>
  : T extends BailTypes
      ? T // bail out on types that shouldn't be unwrapped
      : T extends object ? { [K in keyof T]: UnwrapRef2<T[K]> } : T
```
if `T` is `any[]`, `UnwrapRef<T>` will be `{ [K in keyof T]: UnwrapRef2<T[K]> }` because `any[]` extends `object`.
So I add `Array<any>` to BailType to avoid unwrapping for array type.


wjw99830:
```typescript
interface IMyObject {
   value: string;
   name: string;
}
const myObject = ref<IMyObject>();
```
Here the type-inference works unexpected, type of `myObject` is `Ref<string | null>` not `Ref<IMyObject | null>`, so I add a private field `__ref__` to `class RefImpl` and use it in `UnwrapRef` to avoid conflict with any other objects which have value field.

wjw99830:
Unwrap native Date object will result type-error, avoid it. Because of the private field of RefImpl, return of ref() should also have this field to be allowed to be passed in ref(), so use class Ref instead of interface Ref.

liximomo:
If we do this for `Date`, should we also do the same for `Image`, `HtmlElement`, etc? 
Explicitly specify the type might b a simple and easy solution:

```js
const dateRef: Ref<Date> = ref(new Date());
```

wjw99830:
OK, it seems better. But why are there 10 UnwrapRef ? I try this `...T extends object ? { [K in keyof T]: UnwrapRef<T[K]> }...`, it works fine.

wjw99830:
OK, it seems better.

PR #110, opened 2019-09-01
improve prop typings when using setup(props)

pikax:
Allow getting typings on the props: 

```ts
export default Vue.extend({

  props: {
      myProp: String
  },

  setup(pp, context) {
    const myprop = ref(pp.myProp)
    return {
      myprop
    }
  }
})

```

image

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,12 +1,19 @@
 import Vue, { VueConstructor } from 'vue';
-import { Data, SetupFunction, SetupContext } from './component';
+import { SetupFunction, SetupContext } from './component';
 import { currentVue } from './runtimeContext';
 import { install } from './install';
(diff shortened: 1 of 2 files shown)


pikax:
@liximomo do you agree with this change?

liximomo:
I would advise using `createComponent`.

pikax:
Is there any difference in using `createComponent` and the `2.x object api`? 

Will `createComponent` be the recommended way to use composition-api with typescript (since in 2.x you need to use Vue.extend)?

liximomo:
Yes. You should definitely to use `createComponent` with typescript.

pikax:
Is this PR still useful or just close it?

pikax:
Outdated

PR #127, opened 2019-09-10
add Promise to BailTypes and retain symbolic properties when unwrapping

MattiasMartens:
When wrapping a nested object with `reactive`, I found that some type information was destroyed. Specifically, if a nested property had a symbol field, it would disappear from the inferred return type of the property.

This was triggered by a wrapped Promise. I think Promise should be added to the list of BailTypes because it's not supposed to have any meaningful reactive properties. In general though, the unwrapped versions of objects should retain their symbolic properties, even if it can't unwrap them due to the current limitations of TypeScript.

### DIFF
--- a/src/reactivity/ref.ts
+++ b/src/reactivity/ref.ts
@@ -4,7 +4,7 @@ import { proxy, isPlainObject } from '../utils';
 import { HasDefined } from '../types/basic';
 import { reactive } from './reactive';
 
-type BailTypes = Function | Map<any, any> | Set<any> | WeakMap<any, any> | WeakSet<any>;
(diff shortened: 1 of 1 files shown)


antfu:
Sorry for the delay.

This PR is out of date. We have updated the typing aligning with vue-next:
https://github.com/vuejs/composition-api/blob/70452afe85bcf822b0b6ff94c322fb5f9a9e9a44/src/reactivity/ref.ts#L25-L29

And if you think this is an enhancement to the typing, please make a proposal/PR to vue-next first. Thanks for your contribution. Closing this for now.

Issue #111, opened 2019-09-01
adding object to be reactive transforms the original instance to observable


pikax:
When you add an global object to reactive it changes that instance instead of create an copy of that object.

```javascript
import {ref} from '@vue/composition-api';

const basicRules = [
  { a: 1, b:1},
  { a: 2, b:2},
  { a: 3, b:3},
  { a: 4, b:4},
]

let i = 0
let latestI = -1
setInterval(() => {
  if (i !== latestI) {
    console.log('basicRule:', basicRules[i], basicRules)
    latestI = i
  }
}, 500)


export default {
  name: "App",
  
  setup(){
    const rule = ref(basicRules[0]);

    const c = () => (rule.value = basicRules[i++])

    return {
      rule,
      c
    }
  }
};
```

[try it on codesandbox - open the devtools](https://7buxm.csb.app/)

I would expect the `ref(basicRules[0]);` creating a new object and that object is observable, but instead it changes the actually instance of the object to be observable: 
![image](https://user-images.githubusercontent.com/4620458/64079365-a5eba280-ccde-11e9-8954-259d7d71b204.png)

Is that the expected behavior? if so can it be documented?

Are we required to clone the objects manually?

LinusBorg:
> Is that the expected behavior?

Yes. That's normal Vue 2 behaviour. We never made "copies" of objects before and don't do so here.

In Vue 3 we can use proxies, but we are limited to the way Vue 2 works here.

pikax:
Makes sense.

After reading the [API specs for ref](https://vue-composition-api-rfc.netlify.com/api.html#ref)

> If an object is assigned as a ref's value, the object is made deeply reactive by the reactive method.
when using `ref` is actually the expected behaviour.

Currently `reactive` is behaving the same way as `ref`, I suppose is limited by `Vue.observable`;

Based on the [docs](https://vue-composition-api-rfc.netlify.com/api.html#reactive):

> Takes an object and returns a reactive proxy of the original. This is equivalent to 2.x's Vue.observable().
> ... the returned proxy is not equal to the original object....

Adding to the docs should be enough, I just wasn't sure if that was the expected behaviour or not.

LinusBorg:
We will not add that to the "docs" as those "docs" are the specification of the feature for Vue 3, not the docs for the 2.* plugin.

pikax:
I meant to this library docs on Limitations section, just as you suggested.

PR #143, opened 2019-09-23
docs: added limitation documentation of using reactive

pikax:
Currently using `reactive` will change the original object passed to the function, this is not an expected behaviour, since the [doc](https://vue-composition-api-rfc.netlify.com/api.html#reactive):

> Takes an object and returns a reactive proxy of the original. This is equivalent to 2.x's Vue.observable().
> ... the returned proxy is ***not*** equal to the original object....

But currently when using `reactive` ***will*** change the original object.

***NOTE*** if someone could translate the chinese docs pls

Issue: #111

### DIFF
--- a/README.zh-CN.md
+++ b/README.zh-CN.md
@@ -167,6 +167,11 @@ b.list.push(
 b.list[1].count === 1; // true
 ```
 
+### ***Using*** `reactive` will mutate the origin object
+
+This is an limitation of using Vue.observable in Vue 2.
+> Vue 3 it will return an new proxy object.
+
 ---
 
 ## `watch()` API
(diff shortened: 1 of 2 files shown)


README.zh-CN.md
```diff
@@ -167,6 +167,11 @@ b.list.push(
 b.list[1].count === 1; // true
 ```
 
+### ***Using*** `reactive` will mutate the origin object
```
liximomo: English: ***Using*** `reactive` will mutate the origin object
Chinese: `reactive` 会返回一个修改过的原始的对象

English: This is an limitation of using Vue.observable in Vue 2.
Chinese: 此行为与 Vue 2 中的 `Vue.observable` 一致

English:  > Vue 3 it will return an new proxy object.
Chinese: > Vue 3 中会返回一个新的的代理对象.

PR #145, opened 2019-09-26
Improvement computed type

kahirokunn:


kahirokunn:
Is anyone there?

src/apis/computed.ts
```diff
@@ -9,7 +9,7 @@ interface Option<T> {
 }
 
 // read-only
-export function computed<T>(getter: Option<T>['get']): Ref<T>;
+export function computed<T>(getter: Option<T>['get']): Readonly<Ref<T>>;
```
pikax: I think this is more correct

```suggestion
export function computed<T>(getter: Option<T>['get']): Readonly<Ref<Readonly<T>>>;
```

kahirokunn:
I think so. Thx suggest @pikax !

src/apis/computed.ts
```diff
@@ -9,7 +9,7 @@ interface Option<T> {
 }
 
 // read-only
-export function computed<T>(getter: Option<T>['get']): Readonly<Ref<T>>;
+export function computed<T>(getter: Option<T>['get']): Readonly<Ref<Readonly<T>>>;
```
liximomo: Why we need to type T as Readonly.

pikax: My suggestion might not be correct, based on the rfc, it returns an immutable ref object, I think you shouldn't be allowed to change it, but 2.x allows you to do that.

I actually open an rfc to discuss that: https://github.com/vuejs/rfcs/pull/82

I apologise if I jump into conclusions to early, the [documentation for computed](https://vue-composition-api-rfc.netlify.com/api.html#computed) also changed because of me :/

kahirokunn: Should I to do anything?

src/apis/computed.ts
```diff
@@ -9,7 +9,7 @@ interface Option<T> {
...
 // writable
 export function computed<T>(options: Option<T>): Ref<T>;
```
liximomo: Considering how computed works, the return value of a writable computed should be `Ref<readonly<T>>` instead of Ref<T>.

kahirokunn: Thx review! Yeah, I think so.
So how about this change?
https://github.com/vuejs/composition-api/pull/145/commits/4fbfdcfed3a4fb592b5707d4594f48a25b258149

kahirokunn:
What else can I do?

kahirokunn:
Hello ? 😢

kahirokunn:
Ok !

liximomo:
I prefer this:

```tsx
export function computed<T>(getter: Option<T>['get']): Readonly<Ref<T>>;
export function computed<T>(getter: Option<T>['get']): Ref<T>;
```

kahirokunn:
@liximomo https://github.com/vuejs/composition-api/pull/145#issuecomment-541031680
Sorry, I don't understand this comment. 😢
Can u explain more ? 😢 
Thx 🙏

liximomo:
@kahirokunn 

```tsx
// read-only
export function computed<T>(getter: Option<T>['get']): Readonly<Ref<T>>;
// writable
export function computed<T>(options: Option<T>): Ref<T>;
```

kahirokunn:
I fixed readonly like that.

but I think that's writable type is not better then before.
because writable computed is expected this code.

```typescript
someWritableComputed.value = { hoge: 123 }
```

but that's type allow this code.

```typescript
someWritableComputed.value.hoge = 123
```

kahirokunn:
How do you think about that ?

liximomo:
@kahirokunn You are right. I think `Readonly<Ref<Readonly<T>>>` is better;

kahirokunn:
Thx review @liximomo !
I think we can merge this pull request  if you don't have additional order.

liximomo:
You need to revert this commit https://github.com/vuejs/composition-api/pull/145/commits/d6a6eacb99e9ac63be7e9064489d2fb25bdd370e

kahirokunn:
@liximomo How about now?

Issue #144, opened 2019-09-25
Method context dependent on parentheses use


doman412:
When exporting methods from a compositional function, the `this` context changes depending on if the template uses parentheses to call it. This is not the case with 2.x options methods. I've created a [codepen](https://codepen.io/doman412/pen/VwZREmQ) that demonstrates this bug.

posva:
the methods returned in setup should indeed be bound to the component instance

pikax:
I can fix this, will create a PR in a bit :)

liximomo:
@posva According to the origin RFC:

![image](https://user-images.githubusercontent.com/9816225/65809677-74c59b80-e1d2-11e9-9ae1-7b387268cd75.png)

And what about the consistency of inline render function?

```js
const comp = createComponent({
  setup() {
    const onButton = function() {
      // what would you expect `this` to be?
      console.log(this);
    };

    return () =>
      h('button', {
        on: {
          click: onButton,
        },
      }, 'click me');
  },
});
```

I wouldn't think it should be bound to the component instance.

posva:
You are right. Should we bind to `null` instead to forbid the usage of `this` altogether? We could even use a `Proxy` in development if available to show a warning.

It's inconvenient though, I need to revisit the RFC (I may be forgetting about something we already talked about) because when starting using the Composition API and mixing it with the object api, people will likely need to reference other properties in the component instance. At the same time, some of these properties are not available right away so having access to a partial `this` would lead to even more confusion that not having it all together, so I think disabling `this` altogether seems better. However this would prevent people from changing the `context` of the function later on, which is a limitation that could be avoided

pikax:
Consistent behaviour would be better, since the composition-api behaviour is not aligned with 2.x. 
Tbh with the composition api, you are less likely to access `this` in the function, since most likely you will have through `setup()` closure. 

I think this is still an issue, because is such a nuance behaviour, that will be difficult for someone to spot when doing a mix on `object api` and `composition-api` in the same component, since this package is suppose to allow the new API in 2.x and work along side it.

```js
createComponent({
  template: `<div>
  <p>{{a}}</p>
  <button @click="increment"> increment</button>
  <button @click="increment()"> increment()</button>
</div>`,

  data() {
    return {
      a: 1
    };
  },

  setup() {
    const increment = function() {
      console.log(this)
      this.a++;
    };
    return {
      increment
    };
  }
});
```

I think there's still an issue here, since it should behave the same in this case IMO.

[...]

PR #148, opened 2019-09-27
fix: binding context vm when using function without parentheses 

pikax:
Binding function to the vm context.

Fixes: #144

### DIFF
--- a/test/setup.spec.js
+++ b/test/setup.spec.js
@@ -355,4 +355,46 @@ describe('setup', () => {
       })
       .then(done);
   });
+
+  describe('Bindings', ()=>{
+    it('should bind the vm when calling with parentesis', async ()=>{
+      let context = null;
+      const contextFunction = jest.fn(function (){
+        context = this
+      });
+
+      const vm = new Vue({
+        template: '<div><button @click="contextFunction()"/></div>',
+        setup() {
+          return {
+            contextFunction
+          }
+        }
+      }).$mount();
+  
+      await vm.$el.querySelector('button').click();
+      expect(contextFunction).toBeCalled();
+      expect(context.$el).toBe(vm.$el);
+    });
+
+    it('should bind the vm when calling without parentesis', async ()=>{
+      let context = null;
+      const contextFunction = jest.fn(function (){
+        context = this
+      });
+
+      const vm = new Vue({
+        template: '<div><button @click="contextFunction"/></div>',
+        setup() {
+          return {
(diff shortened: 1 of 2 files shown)


test/setup.spec.js
```diff
@@ -355,4 +355,46 @@ describe('setup', () => {
       })
       .then(done);
   });
+
+  describe('Bindings', ()=>{
```
posva: If you really want to group them, something like `Methods` would be better
```suggestion
  describe('Methods', ()=>{
```

test/setup.spec.js
```diff
@@ -355,4 +355,46 @@ describe('setup', () => {
...
+    it('should bind the vm when calling with parentesis', async ()=>{
```
posva:
```suggestion
    it('binds methods when calling with parenthesis', async ()=>{
```

test/setup.spec.js
```diff
@@ -355,4 +355,46 @@ describe('setup', () => {
...
+          }
+        }
+      }).$mount();
+  
+      await vm.$el.querySelector('button').click();
+      expect(contextFunction).toBeCalled();
+      expect(context.$el).toBe(vm.$el);
+    });
+
+    it('should bind the vm when calling without parentesis', async ()=>{
```
posva:
```suggestion
    it('binds methods when calling without parenthesis', async ()=>{
```

test/setup.spec.js
```diff
@@ -355,4 +355,46 @@ describe('setup', () => {
...
+        setup() {
+          return {
+            contextFunction
+          }
+        }
+      }).$mount();
+  
+      await vm.$el.querySelector('button').click();
+      expect(contextFunction).toBeCalled();
+      expect(context.$el).toBe(vm.$el);
```
posva: this could be moved inside of the `contextFunction` and get rid of the context variable

test/setup.spec.js
```diff
@@ -355,4 +355,46 @@ describe('setup', () => {
...
```
posva: same

test/setup.spec.js
posva: also, shouldn't we just compare to `vm` directly?

pikax: when I created the tests `vm` was being proxied, but now because of the `bind` I can do that 👍

anhoev:
Please merge, i need it because context.slots is not reactive in runtime

Issue #146, opened 2019-09-26
Update v0.3.2 breaks Typescript-only type of props


kukoo1:
Declaring props type with Typescript was broken after updated from v0.3.1 to v0.3.2
```ts
interface Props {
    prop1: string[];
    prop2: number[];
}
export default createComponent({
    props: (['prop1', 'props2'] as unknown) as Props,
 })
```
The compiler complains
` No overload matches this call.
  Overload 1 of 2, '(options: ComponentOptionsWithoutProps<never, unknown>): VueProxy<never, unknown>', gave the following error.
    Type 'Props' is not assignable to type 'undefined'.`

I investigated and found that this change cause the break.
https://github.com/vuejs/composition-api/commit/0565acb4ff56042c811c661893a4c61d100ef33e#diff-f2fa4059dd5efeb563915aed92f20664R81
From `PropsOptions = ComponentPropsOptions` to `PropsOptions extends ComponentPropsOptions = ComponentPropsOptions`

How to exactly declare props type using typescript ?

wjw99830:
This works:
```typescript
interface IProps {
  prop1: string[];
  prop2: number[];
}
export default createComponent<IProps>({
  props: { prop1: Array, prop2: Array },
});
```

pikax:
I would do:

```typescript
import { PropType } from 'vue'
export default createComponent<IProps>({
  props: { prop1: Array as PropType<string[]>, prop2: Array as PropType<number[]>},
});

```

kukoo1:
@wjw99830 That's working ! thanks
@pikax That's working too, fit for those who want to keep values and types in one place.
Note that if you want to declare a prop in an object fashion, you need to use `PropOptions<type, isRequired>` instead.
```ts
props: {
    prop1: {
        type: Array,
        required: true,
        default: ['a'],
    } as PropOptions<string[], true>,
}

**Comment by pikax** [comment]
@kukoo1 actually you don't need, just doing it on the type is enough :)
```typescript
props: {
    prop1: {
        type: Array as PropType<Array[]>,
        required: true,
        default: ['a'],
    },
}
```

mikeyhew:
is there any way to avoid providing the props attribute completely, like it says here? https://github.com/vuejs/rfcs/blob/function-apis/active-rfcs/0000-function-api.md#typescript-only-props-typing

pikax:
That's for v3 only, @vue/composition-api runs on v2

mikeyhew:
[...]

PR #150, opened 2019-09-27
fix: createComponent() should accept tuple props

wjw99830:
[#146](https://github.com/vuejs/composition-api/issues/146#issue-499062781)
Current `createComponent()` couldn't accept tuple props because of a type-error.
Now this works:
```typescript
createComponent({
  props: ['p1', 'p2'] as const,
  setup(props) {
    props.p1 // any
  },
});
```

### DIFF
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -12,7 +12,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [8.x, 10.x, 12.x]
+        node-version: [10.x, 12.x]
 
     steps:
     - uses: actions/checkout@v1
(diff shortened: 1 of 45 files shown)


pikax:
why is `const` needed?

wjw99830:
@pikax
```typescript
type PropKeys<O extends readonly string[]> = O[number];
```
If we don't use `const`, generic type O will be `string` instead of union type.

pikax:
in vue 2.x you can specify with an array without `as const` I think it should be the same here.

wjw99830:
But current `createComponent()` reports an type-error when accept array props and `const` is not required if we don't need type inference or pass a generic argument for props.

antfu:
Close in favor of #364. Thanks for your contribution.

PR #163, opened 2019-10-20
fix(ref): ref(Ref) should return Ref

jo0ger:


src/reactivity/ref.ts
```diff
@@ -117,9 +117,9 @@ export function ref<S, T = unknown, R = HasDefined<S> extends true ? S : RefValu
 ): Ref<R>;
 // implementation
 export function ref(raw?: any): any {
-  // if (isRef(raw)) {
-  //   return {} as any;
-  // }
+  if (isRef(raw)) {
+    return raw;
```
liximomo: We should create a new ref instead of returning `raw`.

LinusBorg: I disagree. In `vue-next`, we also return the raw value:

https://github.com/vuejs/vue-next/blob/16a737dc6717e1cfe039d4fa56b6eacfc75111c0/packages/reactivity/src/ref.ts#L43-L46

antfu:
Sorry for the delay. This is already covered by recent changes.

Closing this for now. Thanks for your contribution.

PR #167, opened 2019-10-24
Support reactive object with 'value' property

### LINKED ISSUE #157: Type infer error if interface has "value" property
define an interface A:
```javasciprt
interface A {
	id:string
	...
	value:string
}
```
if using with **toRefs**
```javascript
toRefs(reactive({a:<A[]>[]}))
```
as an export use function:
```javascript
function useValue(){
	return toRefs(reactive({a:<A[]>[]}))
}
const { a } = useValue()
```

here, type infer is error, **a is to be Ref<string[]>, not be Ref<A[]>**
if A has no property value, it works ok.

### LINKED ISSUE #159: Return type for reactive was wrong
like this:
WX20191018-142502@2x

dvic:
- fixes #157
- fixes #159

### DIFF
--- a/src/reactivity/ref.ts
+++ b/src/reactivity/ref.ts
@@ -14,11 +14,9 @@ export interface Ref<T> {
 // Recursively unwraps nested value bindings.
 // Unfortunately TS cannot do recursive types, but this should be enough for
 // practical use cases...
-export type UnwrapRef<T> = T extends Ref<infer V>
-  ? UnwrapRef2<V>
-  : T extends BailTypes
-      ? T // bail out on types that shouldn't be unwrapped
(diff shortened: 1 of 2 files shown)


dvic:
@liximomo thoughts?

renyamizuno:
So cool! 🆒 

but this pattern fail. (may be)

image

dvic:
@renyamizuno interesting, maybe this could be solved using the new recursive types support in TS, but I'll have to check.

LinusBorg:
I think this will be solved more elegantly with #199 

If you agree I'll close his.

dvic:
agreed, this can be closed

Issue #170, opened 2019-10-30
document on the migration from function-api?


swuecho:
find the original repo is in achive mode, but did not find any doc on how to upgrade

swuecho:
"export 'value' was not found in '@vue/composition-api'

"export 'value' was not found in 'vue'

LinusBorg:
You can read the updated RFC and learn that there's no more `value`, we now have `ref`.

swuecho:
Yes. but some doc alway helps. General user should not have to read the RFC to use a feature. I know there are docs, but can be overwhelming.

antfu:
Closing this for now. Same reason as #176. Thanks for understanding.

PR #176, opened 2019-11-03
migration guide

swuecho:
resolve https://github.com/vuejs/composition-api/issues/170


antfu:
Functional API was something proposal in the very early stage and even before the composition API RFC getting merge. I think it's very out of date and we may not need to put in our README.

Sorry I would close it for now. Thanks for the contribution.

Issue #185, opened 2019-11-11
vue-router doesn't accept components created with createComponent()


masonk:
I pushed a test case to a new repo [here](https://github.com/masonk/vue-router-rtc/commit/b9b4b536783088e719180bb65e828f5e9eb4beaf)

Steps to reproduce this repo:

1. Create a new repo with vue-cli 3. My version was 3.12.1.
2. Accept defaults (router, typescript).
3. `yarn add @vue/composition-api`
4. Edit `Home.vue` to use createComponent()
5. Typescript starts erroring out during type checking:

Error message:
```
    13 | export default createComponent({
    14 |   components: {
ERROR in /home/mason/dev/vue-rtc/vue-rtc/src/router.ts
7:27 Argument of type '{ mode: "history"; base: any; routes: ({ path: string; name: string; component: VueProxy<ComponentPropsOptions<Data>, Data>; } | { path: string; name: string; component: () => Promise<...>; })[]; }' is not assignable to parameter of
 type 'RouterOptions'.
  Types of property 'routes' are incompatible.
    Type '({ path: string; name: string; component: VueProxy<ComponentPropsOptions<Data>, Data>; } | { path: string; name: string; component: () => Promise<typeof import("/home/mason/dev/vue-rtc/vue-rtc/src/views/About.vue")>; })[]' is not assignable to t
ype 'RouteConfig[]'.
      Type '{ path: string; name: string; component: VueProxy<ComponentPropsOptions<Data>, Data>; } | { path: string; name: string; component: () => Promise<typeof import("/home/mason/dev/vue-rtc/vue-rtc/src/views/About.vue")>; }' is not assignable to typ
e 'RouteConfig'.
        Type '{ path: string; name: string; component: VueProxy<ComponentPropsOptions<Data>, Data>; }' is not assignable to type 'RouteConfig'.
          Types of property 'component' are incompatible.
            Type 'VueProxy<ComponentPropsOptions<Data>, Data>' is not assignable to type 'VueConstructor<Vue> | ComponentOptions<Vue, DefaultData<Vue>, DefaultMethods<Vue>, DefaultComputed, PropsDefinition<...>, Record<...>> | AsyncComponentPromise<...> |
 AsyncComponentFactory<...> | undefined'.
[...]

PR #189, opened 2019-11-12
fix: avoid accessing undeclared instance fields on type level

### LINKED ISSUE #187: Instance type of VueProxy allows any field access when props is not specified
version: 0.3.2

### Reproduction

https://codesandbox.io/s/vuecomposition-api-type-issue-dc492

### Expected Behavior

`comp.foo` should be error as there are no field type `foo` in the component.

### Actual Behavior

No error is provided.

---

This behavior would be a problem if the users enable [template type checking which Vetur offers](https://vuejs.github.io/vetur/interpolation.html).
As Vetur uses component instance type to check expression, it cannot provide diagnostic message even if an accessed field is defined on the component.

ktsn:
fix #187 

As `never` is bottom type, `ExtractPropTypes<never>` is resolved as a type that accepts any field access:

```
ExtractPropTypes<never> = {
    readonly [x: string]: undefined;
    readonly [x: number]: undefined;
}
```

which is included in intersections of `ComponentRenderProxy` type, consists of `VueProxy` instance type.

The default type of `Props` seems to be `unknown`. As it is top type, `ExtractPropTypes<unknown>` will be `{}` which is expected type in this case.

### DIFF
--- a/src/component/component.ts
+++ b/src/component/component.ts
@@ -65,15 +65,15 @@ interface ComponentOptionsWithProps<
   setup?: SetupFunction<Props, RawBindings>;
 }
 
-interface ComponentOptionsWithoutProps<Props = never, RawBindings = Data> {
+interface ComponentOptionsWithoutProps<Props = unknown, RawBindings = Data> {
   props?: undefined;
   setup?: SetupFunction<Props, RawBindings>;
 }
 
 // overload 1: object format with no props
 export function createComponent<RawBindings>(
-  options: ComponentOptionsWithoutProps<never, RawBindings>
-): VueProxy<never, RawBindings>;
+  options: ComponentOptionsWithoutProps<unknown, RawBindings>
+): VueProxy<unknown, RawBindings>;
 // overload 2: object format with object props declaration
 // see `ExtractPropTypes` in ./componentProps.ts
 export function createComponent<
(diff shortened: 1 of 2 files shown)


src/component/component.ts
```diff
@@ -22,8 +22,8 @@ export type ComponentRenderProxy<P = {}, S = {}, PublicProps = P> = {
   S;
 
 // for Vetur and TSX support
-type VueConstructorProxy<PropsOptions, RawBindings> = {
-  new (): ComponentRenderProxy<
+type VueConstructorProxy<PropsOptions, RawBindings> = VueConstructor & {
+  new (...args: any[]): ComponentRenderProxy<
```
ktsn: Modified this type as it cannot match existing typings which expects `VueConstructor` like vue-router. By writing argument as `(...args: any[])`, we can mix it with the intersected constructor so that `ComponentRenderProxy` retains even after inherit `VueConstructor`.

https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-2.html

This probably fixes #185 ?

ux-engineer:
Does this get merged?

dodas:
This solves #185 for me!

yoyo930021:
This PR can fix #185.

PR #198, opened 2019-12-01
feat: add `onServerPrefetch`

### LINKED ISSUE #36: lack of life cycle method `serverPrefetch`
There is no life cycle method `serverPrefetch` , so can not be fully used on SSR.

In order to support SSR, the only choice is to remain `serverPrefetch()` method. And then, all methods and data properties used by `serverPrefetch()` has to be remained. And then, very little code can be ported into `setup()` method. And the code is in "mixed style", very ugly.

Akryum:
Closes #36

### DIFF
--- a/README.md
+++ b/README.md
@@ -320,11 +320,11 @@ Even if there is no definitive Vue 3 API for SSR yet, this plugin implements the
 import { onServerPrefetch } from '@vue/composition-api';
 
 export default {
-  setup () {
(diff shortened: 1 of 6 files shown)


galvez:
@Akryum is there any chance you could pass `ssrContext` as first parameter (or even `vm`)?

Akryum:
@galvez Done

galvez:
Thank you!

PR #203, opened 2019-12-15
Add class as an optional property to all components

mehcode:
From what I understand `class` may be passed to all components.

### DIFF
--- a/src/component/componentProps.ts
+++ b/src/component/componentProps.ts
@@ -43,4 +43,6 @@ export type ExtractPropTypes<O, MakeDefaultRequired extends boolean = true> = {
   readonly [K in RequiredKeys<O, MakeDefaultRequired>]: InferPropType<O[K]>;
 } & {
   readonly [K in OptionalKeys<O, MakeDefaultRequired>]?: InferPropType<O[K]>;
+} & {
+  readonly class?: string;
 };


LinusBorg:
Just like any other html attribute. Is there a reason you want to add class's here specifically ?

mehcode:
I'm talking about custom components. Can properties other than class be forwarded implicitly like that? If so, is there some way to indicate all HTML properties then so typescript can type check it's usage? 

I just had a random need to use this on a custom component when I was trying out Typescript.

LinusBorg:
> Can properties other than class be forwarded implicitly like that?

All attributes that you pass to a component and that are not explicitly definied as props in that component are implicitly passed as attributes to the root element.


> If so, is there some way to indicate all HTML properties then so typescript can type check it's usage?

Well that's usually what the shims in the .d.ts files  are doing that come with an new CLI project, for example.

There are also projects like vue-tsx-support that aim to make TSX type inference even more specific

https://www.npmjs.com/package/vue-tsx-support

Either way, since "passing attributes to components" is a general concern and not directly related to this plugin, I don't think that it made sense to add specific attributes here.

LinusBorg:
I close this since I think we agree that adding individual properties to the types isn't the way to go.

PR #237, opened 2020-01-24
type: add typing for setup function to support scoped slots

valentinpalkovic:
Add typings to SetupContext to support scoped slots appropriately.
It is now possible to pass a type into the setupContext generic to increase type
safety.

### DIFF
--- a/src/component/component.ts
+++ b/src/component/component.ts
@@ -40,9 +40,9 @@ type VueProxy<PropsOptions, RawBindings> = Vue2ComponentOptions<
 > &
   VueConstructorProxy<PropsOptions, RawBindings>;
 
-export interface SetupContext {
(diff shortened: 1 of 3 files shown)


valentinpalkovic:
@liximomo PTAL

antfu:
Hi @valentinpalkovic, sorry for the delay.

In this package, we would like to make the typing consistent with Vue 3 as possible as we could. If you think this would be an enhancement for the composition API, please make a proposal/PR in vue-next first. And then we can adapt them to this package. 

https://github.com/vuejs/vue-next/blob/2408a656627358b21aa49209e64d14a1aeec7825/packages/runtime-core/src/component.ts#L121-L125

Sorry I would like to close this for now. Thanks for your contribution.

Issue #231, opened 2020-01-19
setup() return error on null values


ronssij:
Am I missing something or does the Vue 3.x does not allow return on null value variables?

>
    setup () {
     let cookies = Cookies.get('apiUrl') ? Cookies.get('apiUrl') : null
     let state = reactive({
       apiUrl: ''
     })
     
     let test = null

     return {
       state,
       cookies,
       test
     }
    }

> vue.runtime.esm.js?2b0e:619 [Vue warn]: Error in data(): "TypeError: Cannot convert undefined or null to object"

> TypeError: Cannot convert undefined or null to object

jdmerinor:
I got the same issue, I'm pretty sure this has to simply be a mistake

coolhome:
Just tested in vue 3.0 and returning null is not a problem! 

This next part has not been validated yet... Now with vue 2.0 I think I've traced the desired and that the `isReactive` is throwing the error

https://github.com/vuejs/composition-api/blob/3b7eadfbb3868fd2fc67b538ea8ded4af4905832/src/setup.ts#L197

https://github.com/vuejs/composition-api/blob/3b7eadfbb3868fd2fc67b538ea8ded4af4905832/src/reactivity/reactive.ts#L24

This `isReactive` is likely trying to convert the null to object.

https://github.com/vuejs/composition-api/blob/e1b9757d0695643ca940518ada986862c651efba/src/utils.ts#L37-L40

![image](https://user-images.githubusercontent.com/700740/73502604-8c3aae00-4397-11ea-8fea-c86827542981.png)

clementmas:
I had the same problem and the error "TypeError: Cannot convert undefined or null to object" was not very helpful.
Thanks for the PR @coolhome. Hope it gets merged soon.

antfu:
Fixed in v0.6.1

PR #241, opened 2020-01-31
Handle non-reactive nulls (#231)

coolhome:
Fix for #231 - Ensure non reactive nulls are supported.

```
// 'this' cannot be set to null.
Object.hasOwnProperty.call(null, ''); // TypeError Cannot convert undefined or null to object

// Not an Object...
(null).hasOwnProperty('') // TypeError: Cannot read property 'hasOwnProperty' of null
```

Solution would be to check if the obj is null.

### DIFF
--- a/src/reactivity/reactive.ts
+++ b/src/reactivity/reactive.ts
@@ -22,7 +22,7 @@ function isNonReactive(obj: any): boolean {
 
 export function isReactive(obj: any): boolean {
   return (
-    obj !== null &&
+    obj != null &&
     hasOwn(obj, ReactiveIdentifierKey) &&
     obj[ReactiveIdentifierKey] === ReactiveIdentifier
   );
@@ -86,7 +86,7 @@ export function defineAccessControl(target: AnyObject, key: any, val?: any) {
     get: function getterHandler() {
       const value = getter ? getter.call(target) : val;
       // if the key is equal to RefKey, skip the unwrap logic
-      if (key != RefKey && isRef(value)) {
+      if (key !== RefKey && isRef(value)) {
         return value.value;
       } else {
         return value;
--- a/src/reactivity/reactive.ts
+++ b/src/reactivity/reactive.ts
@@ -86,7 +86,7 @@ export function defineAccessControl(target: AnyObject, key: any, val?: any) {
     get: function getterHandler() {
       const value = getter ? getter.call(target) : val;
       // if the key is equal to RefKey, skip the unwrap logic
-      if (key !== RefKey && isRef(value)) {
+      if (key != RefKey && isRef(value)) {
         return value.value;
       } else {
         return value;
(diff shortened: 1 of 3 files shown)


src/utils.ts
```diff
@@ -34,9 +36,8 @@ export function def(obj: Object, key: string, val: any, enumerable?: boolean) {
   });
 }
 
-const hasOwnProperty = Object.prototype.hasOwnProperty;
 export function hasOwn(obj: Object | any[], key: string): boolean {
-  return hasOwnProperty.call(obj, key);
+  return obj !== null && Object.hasOwnProperty.call(obj, key);
```
sodatea: Better to do the `obj !== null` test in the `isReactive` function.
And change the `obj: Object` to `obj: object`, to make the code clearer.

coolhome: @sodatea thanks for the feedback!

src/reactivity/reactive.ts
```diff
@@ -82,7 +86,7 @@ export function defineAccessControl(target: AnyObject, key: any, val?: any) {
     get: function getterHandler() {
       const value = getter ? getter.call(target) : val;
       // if the key is equal to RefKey, skip the unwrap logic
-      if (key !== RefKey && isRef(value)) {
+      if (key != RefKey && isRef(value)) {
```
sodatea: Why change this line?

test/setup.spec.js
```diff
@@ -20,6 +20,28 @@ describe('setup', () => {
...
+  
+  it('should work with non reactive undefined', () => {
+    const vm = new Vue({
+      setup() {
+        return {
+          a: undefined,
+        };
+      },
+    }).$mount();
+    expect(vm.a).toBe(undefined);
```
sodatea: Not a valid test case. Because the `vm.a` would also be `undefined` even when the `setup` function throws.
The alternative is to either test against the log message for warning messages or add an extra non-empty field and see if the value also successfully returns (I personally prefer the latter, as it does not depend on internal implementations).

src/reactivity/reactive.ts
```diff
@@ -21,7 +21,11 @@ function isNonReactive(obj: any): boolean {
 }
 
 export function isReactive(obj: any): boolean {
-  return hasOwn(obj, ReactiveIdentifierKey) && obj[ReactiveIdentifierKey] === ReactiveIdentifier;
+  return (
+    obj !== null &&
```
posva:
```suggestion
    obj != null &&
```

src/reactivity/reactive.ts
posva:
```suggestion
      if (key !== RefKey && isRef(value)) {
```

test/setup.spec.js
antfu, line 42:
```suggestion
      setup() {
        return {
          a: undefined,
          b: 'foobar',
        };
      },
    }).$mount();
    expect(vm.a).toBe(undefined);
    expect(vm.b).toBe('foobar');
```

antfu:
This is covered by #311 and no longer an issue in v0.6.1. However, I think we can take the unit-tests.

PR #261, opened 2020-02-24
types(reactivity): fix corner case on UnwrapRef

yoyo930021:
Same as vuejs/vue-next#614

We need special string or number types sometimes.
Ex. RelativePath, AbsolutePath

```typescript
type AbsolutePath = string & { __brand: unknown }
function readFile (path: AbsolutePath) {
  // do some thing
}
function getAbsolutePath (path: string) {
  if (path.startWith('file://')) return path as AbsolutePath
}
```

I use it for safe use translation keys,
but code is broken when passing UnwrapRef on vetur template interpolation.


LinusBorg:
@ktsn What do you think about this PR?

PR #265, opened 2020-02-26
Grammer

nainardev:
Looks a lot cleaner and made grammer better


LinusBorg:
Thanks for the contribution.

However I think these parts of the README are fine as it is. Especially the install command is clearer without using a shortcut.

PR #276, opened 2020-03-09
fix: the parent setup should be called first

shaonialife:
The parent setup should be called first, otherwise inject cannot get the provide data of the parent class


LinusBorg:
Thanks!

PR #277, opened 2020-03-09
Fix(computed): destroy helper vm of computed to prevent memleak (fix #270)

### LINKED ISSUE #270: computed on a shared reactive causes a memory leak
```html
<!DOCTYPE html>
<html lang="en">
<head>
    <meta charset="UTF-8">
    <title>Title</title>
    <script src="https://cdn.jsdelivr.net/npm/vue"></script>
    <script src="https://unpkg.com/@vue/composition-api/dist/vue-composition-api.umd.js"></script>
</head>
<body>
<div id="app"></div>
<script>
    Vue.use(vueCompositionApi.default);

    //1. Create a globally Shared reactive object
    const store = vueCompositionApi.reactive({
        list: new Array(100000).fill('').map((_, index) => {
            return {
                index
            }
        })
    })

    // 2.Use the global Shared reactive object in the "computed" function in the "Item" component
    const Example = {
        template: `<div>{{ls.length}}</div>`,
        setup() {
            return {
                ls: vueCompositionApi.computed(() => store.list)
            }
        }
    }

    // 3, On the "App" component, use the "toggle" button to continuously create and destroy the "Item" component
    const App = {
        components: {Example},
        template: `
            <div>
                <button @click="show = !show">toggle</button>
                <example v-if="show"></example>
            </div>`,
        setup() {
            return {
                show: vueCompositionApi.ref(false)
            }
        }
    }


    new Vue({
        el: '#app',
        render: h => h(App)
    })
</script>
</body>
</html>
```
Keep clicking the toggle button, you will find that the memory keeps increasing and will not be destroyed.

LinusBorg:
We use a Vue instancefor each computed(), as a helper to actually use Vue's `computed:` option.

However we don't destroy it when the host component is destroyed. When the computed property references reactive state that's external to the host instance, the helper instance stays in memory as it's watching the external state.

Properly destroying witht the host component fixes this.
 
fix #270

### DIFF
--- a/src/apis/computed.ts
+++ b/src/apis/computed.ts
@@ -34,6 +34,8 @@ export function computed<T>(
     },
   });
 
+  vm && vm.$on('hook:destroyed', () => computedHost.$destroy());
+
   return createRef<T>({
     get: () => (computedHost as any).$$state,
     set: (v: T) => {


LinusBorg:
As just discussed on Dicsord, this might break stuff in specific cirtumstances. We should not merge this now and do further investigations.

LinusBorg:
This won't work. It will break any computed refs that are *meant* to persist after their original component has been destroyed.

Will close and investigate other possible ways.

LinusBorg:
Reopened, see here: https://github.com/vuejs/composition-api/issues/270#issuecomment-621200581

commyfriend:
It seems this pr has been waiting for a long time to be merged.

LinusBorg:
Yeah I think it's fine as well @antfu

PR #278, opened 2020-03-09
type: Added Element type to ref BailTypes

kedrzu:
Suppose I want to return ref to the template to be filled with my html element:
`const element = ref<HTMLElement>(null);`

Then I wrap it with reactive:
```
const state = reactive({
    element,
    ...someOtherProps
})
```

Element goes through `UnwrapRef` and its type inference is broken.


LinusBorg:
@ktsn I think this makes sense, need a second opinion though.

PR #283, opened 2020-03-16
fix: isReactive & isNonReactive throws on null value

kedrzu:
suppose you have a component: 

```
{
    setup() {
        return {
            foo: null
        }
    }
}
```

If your property is `null` or `undefined` a function `isReactive()` throws error, because it tries to call `hasOwnProperty()` on `null` value.

I suppose the same case is for the `isNonReactive()` function, so I made fixes in both.

Some call stack for the reference:
image


kedrzu:
I see there is already a fix for that

PR #298, opened 2020-03-30
fix(watch): component should runCleanup() on unmount

### LINKED ISSUE #293: WatchEffect's onCleanup is not firing when component is unmounted
Greetings,

I am using `watchEffect` in a component that is loaded by `router-view`.
After the route is changed, this component is unmounted.
I was expecting `onCleanup` to fire at this time.
Is this a correct expectation?

To my humble information, `vue-router` does not cache routing components, so I am assuming they are being destroyed. If that is not the case, then how can I ensure proper cleanup while leveraging the router's cache?

Btw., the `Vue Composition API` official documentation is calling onCleanup as onInvalidate which was a bit confusing to me. May be I was on the wrong docs version.

Thank you.

LinusBorg:
* watchEffect now  properly calls `runCleanup()` when the component unmounts.
* I applied the same change for "normal" watchers as I found that the bug applied to those as well.

Would like a second pair of eyes though, to verify.

close #293

### DIFF
--- a/src/apis/watch.ts
+++ b/src/apis/watch.ts
@@ -139,6 +139,16 @@ function createVueWatcher(
   return vm._watchers[index];
 }
 
+// We have to monkeypatch the teardown function so Vue will run
+// runCleanup() when it tears down the watcher on unmmount.
+function patchWatcherTeardown(watcher: VueWatcher, runCleanup: () => void) {
+  const _teardown = watcher.teardown;
+  watcher.teardown = function(...args) {
+    _teardown.apply(watcher, args);
+    runCleanup();
+  };
+}
+
 function createWatcher(
   vm: ComponentInstance,
   source: WatcherSource<unknown> | WatcherSource<unknown>[] | SimpleEffect,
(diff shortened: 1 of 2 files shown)


src/apis/watch.ts
```diff
@@ -188,8 +198,13 @@ function createWatcher(
       before: runCleanup,
     });
 
+    patchWatcherTeardown(watcher, runCleanup);
+
     // enable the watcher update
     watcher.lazy = false;
+    if (vm !== fallbackVM) {
+      vm._watchers.push(watcher);
```
pikax, line 206: Why are you pushing the watch again to the vm?
```ts
// vm._watchers[0] === vm._watchers[1]: true
```
[createVueWatcher](https://github.com/vuejs/composition-api/blob/fix-293-watch_cleanup-on-unmount/src/apis/watch.ts#L117-L140) already pushes the watcher to `vm._watchers` and returns the last

LinusBorg: good catch, will remove that.

LinusBorg: done

PR #300, opened 2020-04-01
fix(watch): update watch options from lazy to immediate

### LINKED ISSUE #266: Adjust watch behaviour to match latest RFC update
The documentation for watch states:

>  It also is lazy by default - i.e. the callback is only called when the watched source has changed.

However, when I use it, I see that the callback is called even if the data I watch doesn't change after I called watch. If I set the option `{ lazy: true }`, the callback is not called before the data change.

I also think the documentation should state that this is the same option as `immediate` in 
Vue 2. Unless there are any differences? If so, they should be stated.

antfu:
This aligns with the latest RFC: https://vue-composition-api-rfc.netlify.com/api.html#watch

`lazy` is renamed to `immediate` and default to false. 

This is a BREAKING CHANGE. Not sure if we should make backward compatible and raise warnings.

Resolves #266

### DIFF
--- a/src/apis/watch.ts
+++ b/src/apis/watch.ts
@@ -21,10 +21,16 @@ type MapSources<T> = {
 
 type FlushMode = 'pre' | 'post' | 'sync';
 
-interface WatcherOption {
-  lazy: boolean; // whether or not to delay callcack invoking
-  deep: boolean;
+interface WatchEffectOptions {
   flush: FlushMode;
+  // Not supported options:
+  // onTrack?: (event: DebuggerEvent) => void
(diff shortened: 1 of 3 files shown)


antfu:
Any update on this? My lib [vueuse](https://github.com/antfu/vueuse) is to provide utils for both Vue 2 and 3. This inconsistent makes me hard to support both through the same API.

Is there any plan to merge this PR? Let me know if there are any changes needed. Thanks.

zgayjjf:
Is there any plan to merge or improve this PR?
The difference between RFC and this package confused a lot users.

antfu:
Closing this as it's already being implemented in #311.

PR #301, opened 2020-04-02
fix: avoid including Function type in inferred props type

### LINKED ISSUE #291: Typescript: 'Function' type of PropType introduced in 0.5.0 is annoying when using Vetur
When using Vetur's template interpolation service, object props cannot be used directly in the template from 0.5.0, without getting syntax errors:

ss

Exposing the user through setup, i.e.
```ts
  props: {
    user: Object as PropType<IUser>,
  },
  setup(props) {
    return { user: props.user as IUser }
  }
});
```
works for Vetur, but then the console will print warnings:

```
[Vue warn]: The setup binding property "user" is already declared as a prop.
```

ktsn:
fix #291 

The issue was caused by [distribution of conditional types](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-8.html#distributive-conditional-types). The reason there is no issue when we declare props type via `{ type: String }` format is `T extends { type: PropType<infer V, boolean> }` won't be distributed.

The solution is just avoiding distribution in the conditional types in `InferPropType`. To do that, I wrapped all conditional types with tuple.

### DIFF
--- a/test/types/defineComponent.spec.ts
+++ b/test/types/defineComponent.spec.ts
@@ -1,4 +1,11 @@
-import { createComponent, defineComponent, createElement as h, ref, SetupContext, PropType } from '../../src';
+import {
+  createComponent,
+  defineComponent,
+  createElement as h,
+  ref,
+  SetupContext,
+  PropType,
+} from '../../src';
 import Router from 'vue-router';
 
 const Vue = require('vue/dist/vue.common.js');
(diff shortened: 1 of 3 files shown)


test/types/defineComponent.spec.ts
```diff
@@ -105,6 +112,24 @@ describe('defineComponent', () => {
     expect.assertions(3);
   });
 
+  it('custom props type inferred from PropType', () => {
```
pikax, line 115: this test is passing on `master` branch  🤔

ktsn: I guess this is because the behavior prior TS 3.9 because `PropsType` includes `{}` type which can be assigned to `{ user?: User }`. 🤔 

https://devblogs.microsoft.com/typescript/announcing-typescript-3-9-rc/
> Generally, an intersection type like A & B is assignable to C if either A or B is assignable to C

Once I updated TS to 3.9, the test fails on master.

pikax:
@antfu can we get this merged before release of `0.6.0`?

Bug is still in master:
```ts

interface A {
  a: String
}

defineComponent({
  props: {
    a: Object as PropType<A>
  },
  setup(props){
    props.a
  }
})
```

antfu:
This does not seems to fix for me:

image
image

Meanwhile, vue-next does not use this hack while solving the issue. I guess we need to rework the props typing to align with vue-next instead. And since the release has been delayed for too long, my plan would be delivering the release first and delay this PR into the next release so we can have more time to better take care of it.

antfu:
Closing this in favor of #352. Thanks.

PR #304, opened 2020-04-06
Set default lazy option to true in watch api

### LINKED ISSUE #266: Adjust watch behaviour to match latest RFC update

takumi-n:
This pull request fixes that `watch` function runs when a source variable has been defined.

https://vue-composition-api-rfc.netlify.com/api.html#watch
> The watch API is the exact equivalent of the 2.x this.$watch (and the corresponding watch option). watch requires watching a specific data source, and applies side effects in a separate callback function. It also is lazy by default - i.e. the callback is only called when the watched source has changed.

As stated in Vue Composition API documents, the callback function passed to `watch` should run only when the source has changed.

vuejs/composition-api always sets `lazy` field of `WatcherOption` to `false` by default,
https://github.com/vuejs/composition-api/blob/master/src/apis/watch.ts#L60

but it should be `true` in `watch`

Resolve #266

### DIFF
--- a/src/apis/watch.ts
+++ b/src/apis/watch.ts
@@ -291,7 +291,7 @@ export function watch(
     callback = null;
   }
 
-  const opts = getWatcherOption(options);
+  const opts = getWatcherOption({ lazy: true, ...options });
   const vm = getWatcherVM();
 
   return createWatcher(vm, source, callback, opts);
--- a/test/apis/state.spec.js
+++ b/test/apis/state.spec.js
@@ -27,7 +27,7 @@ describe('api/ref', () => {
     watch(a, () => {
       dummy = a.value;
     });
-    expect(dummy).toBe(1);
+    expect(dummy).toBeUndefined();
     a.value = 2;
     waitForUpdate(() => {
       expect(dummy).toBe(2);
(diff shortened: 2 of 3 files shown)


antfu:
Hi, I think the `lazy` option is replaced by `immediate` and default to false. I have already made PR #300 addressing that.

takumi-n:
@antfu 
Hi, I have overlooked your PR. 
I think that vuejs/composition-api keeps `lazy` for two reasons.

1. As you have said, switching from `lazy` to `immediate` breaks compatibility.
2. `vuejs/composition-api` could differ from RFC and Vue.js 3 because `lazy` is stated in RFC

jacekkarczmarczyk:
https://github.com/vuejs/vue-next/blob/master/CHANGELOG.md#breaking-changes-5 @takumi-n

takumi-n:
@jacekkarczmarczyk 
Thanks. I didn't know that at all.

I think #300 is better.

PR #308, opened 2020-04-18
Update README

lcdonaldson:
Without stating that these imports need to be used in the main.js file, users might struggle to know where they go. It took me a while of experimentation and looking around to figure it out. It might be nice to show the difference from a vue 2 file to this new change for an example as well, rather than just saying install.


posva:
Thanks but this is unfortunately not enough and can be misleading depending on the context the developer is trying to add the plugin (plain html, vue-cli, nuxt, other).

PR #309, opened 2020-04-19
feat: add unref API to match API with vue3

pikax:
Same implementation as vue-next 

https://github.com/vuejs/vue-next/blob/a51b05267231fb1770d6fc727bc276475c21d27f/packages/reactivity/src/ref.ts#L69-L71

```ts
export function unref<T>(ref: T): T extends Ref<infer V> ? V : T {
  return isRef(ref) ? (ref.value as any) : ref
}
```


pikax:
merged into https://github.com/vuejs/composition-api/pull/309

wangsongiam:
can we make a release for this API? It's fairly useful.

PR #310, opened 2020-04-19
types: making ComputedRef types inline with vue3 and provide ComputdRef type

pikax:
Making computed return types inline with the vue 3

```ts
computed(() => ({ a: 1 })).value = 2 as any // type error :correct:
computed(() => ({ a: 1 })).value.a = 2 // works
computed<{ a: number }>(
  {
    get() {
      return { a: 1 }
    },
    set(v) {}
  }
).value.a = 2 // workd
```

### DIFF
--- a/src/apis/computed.ts
+++ b/src/apis/computed.ts
@@ -8,14 +8,20 @@ interface Option<T> {
   set: (value: T) => void;
 }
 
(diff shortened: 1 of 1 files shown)


pikax:
merged into https://github.com/vuejs/composition-api/pull/309

PR #311, opened 2020-04-19
feat: adding missing reactivity api from vue-next

pikax:
This plugin is falling behind on the reactivity API 

This PR is aimed to make the reactivity API close as possible to vue-next.

## Current problems

- `effect` not ported
- `readonly` not ported
- `makeRaw` this is quite cheat, since the original object gets change, having this doesn't do much
- `reactive([])` - because of the v2 reactivity system arrays cannot be reactive, should be warn?

## Breaking
- Watch effect now follows the same behaviour as v3

## Added
- `shallowReactive`
- `shallowRef`

## Changes
- `reactive` marks all possible objects as reactive to conform to vue 3
- `set` marks reactive


LinusBorg:
This looks like great work!

I'll need a bit more time to properly review this. I'm on it tough!

src/reactivity/reactive.ts
```diff
@@ -123,38 +125,136 @@ function observe<T>(obj: T): T {
...
+    return;
+  }
+
+  if (!isPlainObject(obj) || isReactive(obj) || isRaw(obj) || !Object.isExtensible(obj)) {
+    return obj as any;
+  }
+
+  const observed = observe({});
+  // def(obj, ReactiveIdentifierKey, ReactiveIdentifier);
+  markReactive(observe, true);
```
LinusBorg, line 142: shouldn't that be `markReactive(observed, true)` ?

pikax: yes 😅

src/apis/watch.ts
```diff
@@ -1,30 +1,46 @@
...
+type MapOldSources<T, Immediate> = {
+  [K in keyof T]: T[K] extends WatchSource<infer V>
+    ? Immediate extends true
+      ? (V | undefined)
+      : V
+    : never;
 };
 
+export interface WatchOptionsBase {
+  flush?: 'pre' | 'post' | 'sync';
```
LinusBorg, line 32: Could be replace with `flush?: Flushmode`

LinusBorg:
So I have done a review.

- changes to reactivity look fine
- tests: We now have a lot of tests where we switched to { immediate: true } (since that's how watch behaves now) and seemingly very little are left testing the now-default `{immediate: true }` behaviour. But we still have everything covered, I think. just feels weird to have a majority of the tests use the non-default.
- Types: It's awesome that you took the extra steps and renamed everything to be in line with Vue 3, that looks like it was a lot of work and is highly apprechiated. For the really complex types that you ported from the Vue 3 codebase I can say little as they are a bit over my head for the most part 😅

So from m< perspective this looks fine!

pikax:
I tried not to make too many changes to the current tests, focused more on porting the tests from `vue-next`.

Hopefully this types will improve, and they were basically copy&paste from `vue-next` so updating them in the future should be straight forward, I'm also thinking in bringing the typescript type tests over, but waiting for the typescript 3.9 PR to be merged and the `mixins`

pikax:
@antfu sorry missed your comment, didn't noticed the errors! Thank you

test/v3/runtime-core/apiLifecycle.spec.ts
```diff
@@ -0,0 +1,354 @@
...
+    };
+    const vm = new Vue(Comp).$mount();
+    //render(h(Comp), root);
+
+    count.value++;
+    await nextTick();
+    expect(fn).toHaveBeenCalledTimes(1);
+  });
+
+  // it('onBeforeUnmount', async () => {
```
pikax, line 98: Comment this test because couldn't make this work, this was brought from `vue-next` not sure if we should fix this

test/v3/runtime-core/apiLifecycle.spec.ts
```diff
@@ -0,0 +1,354 @@
...
+
+  //   new Vue(Comp).$mount(root);
+  //   //render(h(Comp), root);
+
+  //   toggle.value = false;
+  //   await nextTick();
+  //   expect(fn).toHaveBeenCalledTimes(1);
+  // });
+
+  // it('onUnmounted', async () => {
```
pikax, line 127: ditto

PR #344, opened 2020-05-31
fix(type): fix incorrect ref unwrap

### LINKED ISSUE #257: Unexpected type inference when creating ref in function call
I'm getting unexpected type inference when creating a `ref` in a function call. The `ref('')` is interpreted as `ref<number>` which could definitely cause bugs (this actually happened to me as I was writing a unit test). The simple example below should provide all of the details but please let me know if there is anything else you need!

```typescript
const foo = (bar: Ref<number>) => {
  return bar;
};

// there are no TS errors in any of this code
// ref('') -> ref<unknown, string, number>(raw: string): Ref<number>
const result = foo(ref('')); // result -> Ref<number>
```

Thanks for your help!

antfu:
resolve #257

Align with vue-next
https://github.com/vuejs/vue-next/blob/c6217b4d46cbd3d9e250f2f84de4846d3484c42d/packages/reactivity/src/ref.ts#L30-L37

~I am not sure how these lines work in our current codebase~
```ts
// with init value: a = ref({ a: ref(0) })
// typeof a will be Ref<{ a: number }>
export function ref<S, T = unknown, R = HasDefined<S> extends true ? S : RefValue<T>>(
  raw: T
): Ref<R>;
```
~I kept it for now. But since it's not in vue-next, should we just remove it in order to have better type consistency with vue-next?~

**Edit**: Removed and it's exactly matched to vue-next now.

### DIFF
--- a/src/reactivity/ref.ts
+++ b/src/reactivity/ref.ts
@@ -12,7 +12,7 @@ type BailTypes = Function | Map<any, any> | Set<any> | WeakMap<any, any> | WeakS
 type BaseTypes = string | number | boolean;
 
 declare const _refBrand: unique symbol;
-export interface Ref<T> {
+export interface Ref<T = any> {
   readonly [_refBrand]: true;
   value: T;
 }
--- a/src/reactivity/ref.ts
+++ b/src/reactivity/ref.ts
@@ -1,7 +1,6 @@
 import { Data } from '../component';
 import { RefKey } from '../symbols';
 import { proxy, isPlainObject, warn } from '../utils';
-import { HasDefined } from '../types/basic';
 import { reactive, isReactive, shallowReactive } from './reactive';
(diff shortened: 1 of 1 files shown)


pikax:
Once https://github.com/vuejs/composition-api/pull/311 is merged we should be able keep the `vue-next` types in sync easily

antfu:
@pikax This seems to not be covered by #311. I have made it exactly matched to vue-next now.

package.json
```diff
@@ -42,6 +42,7 @@
     "cross-env": "^7.0.2",
     "husky": "^4.2.5",
     "jest": "^26.0.1",
+    "jest-cli": "^26.0.1",
```
pikax, line 45: Is jest cli needed?

antfu: You are right, my mistake

PR #352, opened 2020-06-07
types: Remove Function type from PropType type

### LINKED ISSUE #291: Typescript: 'Function' type of PropType introduced in 0.5.0 is annoying when using Vetur

pikax:
fix #291 

Gets the correct type for the prop

Similar to https://github.com/vuejs/composition-api/pull/301

```ts
interface A {
  a: String
}

defineComponent({
  props: {
    a: Object as PropType<A>
  },
  setup(props){
    props.a 
  }
})
```

### DIFF
--- a/src/component/componentProps.ts
+++ b/src/component/componentProps.ts
@@ -30,6 +30,16 @@ type RequiredKeys<T, MakeDefaultRequired> = {
 
 type OptionalKeys<T, MakeDefaultRequired> = Exclude<keyof T, RequiredKeys<T, MakeDefaultRequired>>;
 
+type ExtractFunctionPropType<
+  T extends Function,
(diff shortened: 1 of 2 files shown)


antfu:
I have double-checked the code and I think it's alright, let's ship it with v0.6.0!

PR #360, opened 2020-06-09
fix(lifecycle):fix memory leak cause by incorrect lifecycle merge

shaonialife:
current ```injectHookOption``` will cause memory leak, change it to like this
```typescript
function injectHookOption(
  Vue: VueConstructor,
  vm: ComponentInstance,
  hook: string,
  val: Function,
  customHook?: boolean
) {
  if (!customHook) {
    vm.$on(`hook:${hook}`, wrapHookCall(vm, val));
  } else {
    const hookMethods = (vm as any)[HookMethodsKey] || ((vm as any)[HookMethodsKey] = {});
    const mergeFn = Vue.config.optionMergeStrategies[hook];
    hookMethods[hook] = mergeFn(hookMethods[hook], wrapHookCall(vm, val));
  }
}
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+# 0.5.0
+
+- New: `watchEffect` function, lingin up with the latest version of the RFC ([RFC docs](https://vue-composition-api-rfc.netlify.com/api.html#watcheffect)) (#275)
+- Fix: `setup` from a mixin should called before the component's own (#276)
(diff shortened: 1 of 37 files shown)


antfu:
@shaonialife Can you show me some reproduction that causing the mem leak with the current version? Thanks.

shaonialife:
Sorry, due to my careless testing, I found that another bug caused the memory leak. I will open a new issue to describe this problem

shaonialife:
#365

PR #361, opened 2020-06-09
fix(reactivity): unwrap nested refs on the template

pikax:
### Breaking change

This unwraps nested refs on the template

NOTE: this needs to be intensively tested

### DIFF
--- a/src/reactivity/unwrap.ts
+++ b/src/reactivity/unwrap.ts
@@ -1,45 +1,51 @@
-import { isRef } from './ref';
-import { proxy, isFunction, isObject, isArray } from '../utils';
-import { isReactive } from './reactive';
+import { isRef } from './ref'
+import { proxy, isFunction, isObject, isArray } from '../utils'
+import { isReactive } from './reactive'
 
 export function unwrapRefProxy(value: any) {
   if (isFunction(value)) {
-    return value;
+    return value
   }
 
   if (isRef(value)) {
-    return value;
+    return value
   }
 
   if (isArray(value)) {
-    return value;
+    return value
   }
 
   if (isReactive(value)) {
-    return value;
+    return value
   }
 
   if (!isObject(value)) {
-    return value;
+    return value
   }
 
-  const obj: any = {};
+  if (!Object.isExtensible(value)) {
+    return value
+  }
+
+  const obj: any = {}
 
   // copy symbols over
-  Object.getOwnPropertySymbols(value).forEach((s) => (obj[s] = (value as any)[s]));
+  Object.getOwnPropertySymbols(value).forEach(
+    (s) => (obj[s] = (value as any)[s])
+  )
 
   for (const k of Object.keys(value)) {
-    const r = value[k];
+    const r = value[k]
     // if is a ref, create a proxy to retrieve the value,
     if (isRef(r)) {
-      const set = (v: any) => (r.value = v);
-      const get = () => r.value;
+      const set = (v: any) => (r.value = v)
+      const get = () => r.value
 
-      proxy(obj, k, { get, set });
+      proxy(obj, k, { get, set })
     } else {
-      obj[k] = unwrapRefProxy(r);
+      obj[k] = unwrapRefProxy(r)
     }
   }
 
-  return obj;
+  return obj
 }
--- a/src/reactivity/unwrap.ts
+++ b/src/reactivity/unwrap.ts
@@ -1,0 +1,46 @@
+import { isRef } from './ref';
+import { proxy, isFunction, isObject, isArray } from '../utils';
+import { isReactive } from './reactive';
+
+export function unwrapRefProxy(value: any) {
+  if (isFunction(value)) {
+    return value;
+  }
+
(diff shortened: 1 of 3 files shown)


test/setup.spec.js
```diff
@@ -382,45 +382,86 @@ describe('setup', () => {
       .then(done);
   });
 
+  it('should unwrap on the template', () => {
+    const vm = new Vue({
+      setup() {
+        const r = ref('r');
+        const nested = {
+          a: ref('a'),
+        };
```
antfu, line 391: Maybe we could cover deep nested as well.

pikax: Added a few more, can you check that's good?

src/reactivity/unwrap.ts
```diff
@@ -0,0 +1,46 @@
...
+    const v = value[k];
+    let r = unwrapRefProxy(v);
+    // if is a ref, create a proxy to retrieve the value,
+    if (isRef(r)) {
+      const set = (v: any) => (value[k].value = v);
+      const get = () => value[k].value;
+
+      proxy(obj, k, { get, set });
+    } else {
+      obj[k] = r;
```
antfu, line 41:
```suggestion
    const r = value[k];
    // if is a ref, create a proxy to retrieve the value,
    if (isRef(r)) {
      const set = (v: any) => (r.value = v);
      const get = () => r.value;

      proxy(obj, k, { get, set });
    } else {
      obj[k] = unwrapRefProxy(r);
```

Not an issue, but I think this could make the code a bit cleaner.

src/reactivity/unwrap.ts
```diff
@@ -0,0 +1,45 @@
...
+  }
+
+  if (isReactive(value)) {
+    return value;
+  }
+
+  if (!isObject(value)) {
+    return value;
+  }
+
```
antfu, line 25: Should we add `!Object.isExtensible(value)`?

pikax: Can you provide a case? 

I don't think so, because we don't alter the object, just create a clone of it and unwrap `refs`

antfu: I think a frozen object should remain untouched (and it will disable Vue's reactivity interpretation inside it, which can improve performance in some applications).

We also run the unwrap before `markRaw`, if we do clone it, #366 will never be covered in this usage.

pikax: added the condition

antfu:
LGTM. If you are ok with it, I can do the resolving and merge it now.

pikax:
Sure :)

src/reactivity/unwrap.ts
```diff
@@ -0,0 +1,51 @@
...
+  if (isRef(value)) {
+    return value
+  }
+
+  if (isArray(value)) {
+    return value
+  }
+
+  if (isReactive(value)) {
```
carsoli, line 18: if you don't mind answering, I would really be grateful if you take the time

seems to me, that we have a base-case if `value` is `reactive`, and we don't recursively unwrap

but on the test-cases it doesn't seem to be the case:
https://github.com/vuejs/composition-api/pull/361/files#diff-ffb3fccc5cc98e58adb1477f88e460b83b048c4b693b3fdb479a559f0c66cb88R484-R490

https://github.com/vuejs/composition-api/pull/361/files#diff-ffb3fccc5cc98e58adb1477f88e460b83b048c4b693b3fdb479a559f0c66cb88R550-R552

asking because i'm running into a reproduction and for a minute i was tricked by a premature understanding of this check

pikax: `reactive` has some caveats on [Ref Unwrapping](https://v3.vuejs.org/guide/reactivity-fundamentals.html#ref-unwrapping)

but if a value is reactive, you can safely assume that it's already unwrapping.

This commit is pretty old, some things changed since them

test/setup.spec.js
```diff
@@ -466,6 +467,107 @@ describe('setup', () => {
...
+        <p id="nested_aa_bb_cc">{{ nested.aa.bb.cc }}</p>
+
+        <p id="nested_aaa_b">{{ nested.aaa.b }}</p>
+        <p id="nested_aaa_bb_c">{{ nested.aaa.bb.c }}</p>
+        <p id="nested_aaa_bb_cc">{{ nested.aaa.bb.cc }}</p>
+
+        <p id="nested_aaaa_b">{{ nested.aaaa.b }}</p>
+        <p id="nested_aaaa_bb_c">{{ nested.aaaa.bb }}</p>
+        <p id="nested_aaaa_bbb_cc">{{ nested.aaaa.bbb.c }}</p>
+        <p id="nested_aaaa_bbb_cc">{{ nested.aaaa.bbb.cc }}</p>
```
carsoli, line 530: same id, different ref, and confused with :528 which shouldn't unwrap - the tests for those 3 cases are wrong (overwriting one another)

antfu: Nice catch. Would you mind open a PR? Thanks.

PR #347, opened 2020-06-02
feat(test): add dts testing

antfu:
Added some of the dts testing (ref and watch) copied from vue-next. Will try to adopting more in the future.

This PR also updates Github Actions to using yarn and triggers dts tests.

PR #364, opened 2020-06-09
fix: defineComponent() with array props

antfu:
This continues #150, as it's kinda out-of-date. The type declarations are ported from vue-next.

Before
image

After
image

### DIFF
--- a/test/types/defineComponent.spec.ts
+++ b/test/types/defineComponent.spec.ts
@@ -144,6 +144,17 @@ describe('defineComponent', () => {
     expect.assertions(2);
   });
 
+  it('should accept tuple props', () => {
+    const App = createComponent({
+      props: ['p1', 'p2'],
+      setup(props) {
+        props.p1;
+        props.p2;
+      },
+    });
+    new Vue(App);
+  });
+
   it('infer the required prop', () => {
     const App = defineComponent({
       props: {
--- a/test/types/defineComponent.spec.ts
+++ b/test/types/defineComponent.spec.ts
@@ -145,14 +145,15 @@ describe('defineComponent', () => {
   });
 
   it('should accept tuple props', () => {
-    const App = createComponent({
+    const App = defineComponent({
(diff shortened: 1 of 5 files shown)


antfu:
Tests passed on my machine but failed on Github Actions. I guess it's because it uses npm without package-lock.json which may cause dependencies inconsistency. Hold to see if the error still persists after #347 gets merged.

antfu:
Passed with ts `v3.9.4` but failed with the latest `v3.9.5`. Let's hold it for a while.

pikax:
Updated to `3.9.5` on `vue-next` and no error, do you want me to fix this?

antfu:
@pikax it's might due to that I have moved the array overload in front of the object one. And yes, it would be great if you can help out :)

pikax:
Replace `PropOptions` with: 
```ts
type DefaultFactory<T> = () => T | null | undefined;

interface PropOptions<T = any> {
  type?: PropType<T> | true | null;
  required?: boolean;
  default?: T | DefaultFactory<T> | null | undefined;
  validator?(value: unknown): boolean;
}
```

test/types/defineComponent.spec.ts
```diff
@@ -144,6 +144,17 @@ describe('defineComponent', () => {
     expect.assertions(2);
   });
 
+  it('should accept tuple props', () => {
+    const App = createComponent({
```
pikax, line 148: Do you mean `defineComponemt`?

antfu: Yes, just fixed. I also updated the type of createComponent

antfu:
It works. Thanks!

FloryanFilip:
Can I somehow define my prop as tuple value?
`			plannerDateRange: { type: Array as PropType<[string, string]>, required: true },
`
The above produces compilation error: 
Screenshot 2021-01-19 at 14 29 28

PR #369, opened 2020-06-10
Create LICENSE

antfu:


antfu:
Would the name field alright? @LinusBorg @liximomo

LinusBorg:
Good question. liximomo is definitely the original author, but I doubt that a github handle can be used in a LICENSE

antfu:
I have checked his another project's license: https://github.com/liximomo/vscode-sftp/blob/master/LICENSE
Or should I make this one exactly the same as the one in `vscode-sftp`?

Issue #379, opened 2020-06-12
"Maximum call stack size exceeded" in unwrapRefProxy


sapphi-red:
```
RangeError: Maximum call stack size exceeded
    at unwrapRefProxy (vue-composition-api.module.js?750b:657)
    at _loop_1 (vue-composition-api.module.js?750b:648)
    at unwrapRefProxy (vue-composition-api.module.js?750b:654)
    at _loop_1 (vue-composition-api.module.js?750b:648)
    at unwrapRefProxy (vue-composition-api.module.js?750b:654)
    at _loop_1 (vue-composition-api.module.js?750b:648)
    at unwrapRefProxy (vue-composition-api.module.js?750b:654)
    at _loop_1 (vue-composition-api.module.js?750b:648)
    at unwrapRefProxy (vue-composition-api.module.js?750b:654)
    at _loop_1 (vue-composition-api.module.js?750b:648)
```

https://github.com/vuejs/composition-api/blob/8c449c1a2494c4fdad51490073167766f9cd38af/src/reactivity/unwrap.ts#L5-L38
Changing this func like below will fix this error. (refs #361)

```ts
export function unwrapRefProxy(value: any) {
  return value
}
```

antfu:
Looks like the object you passed to setup has cycle. You may also want to make sure if it's expected.

/cc @pikax should we keep a "traveled list"?

sapphi-red:
Oh, exactly it was so. (checked by passing object to `JSON.stringify`)

```
TypeError: Converting circular structure to JSON
    --> starting at object with constructor 'Vue'
    |     property '$options' -> object with constructor 'Object'
    |     property 'router' -> object with constructor 'VueRouter'
    --- property 'app' closes the circle
```

sapphi-red:
This object maybe `SetupContext`.

pikax:
I will PR it in a bit

PR #380, opened 2020-06-12
fix(setup): call stack exceeded when returning circular dependency

pikax:
fix https://github.com/vuejs/composition-api/issues/379

### DIFF
--- a/src/reactivity/unwrap.ts
+++ b/src/reactivity/unwrap.ts
@@ -2,21 +2,25 @@ import { isRef } from './ref'
 import { proxy, isFunction, isPlainObject, isArray } from '../utils'
 import { isReactive } from './reactive'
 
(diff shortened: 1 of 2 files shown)


src/reactivity/unwrap.ts
```diff
@@ -2,21 +2,25 @@ import { isRef } from './ref'
 import { proxy, isFunction, isPlainObject, isArray } from '../utils'
 import { isReactive } from './reactive'
 
-export function unwrapRefProxy(value: any, processed = new Set<any>()) {
+export function unwrapRefProxy(value: any, map = new WeakMap()) {
+  let processed = map.get(value)
+  if (processed) {
```
antfu, line 7: This will bypass falsy values. `map.has` would be better here.

Issue #151, opened 2019-10-02
Failed to mount when returning render function from setup() while testing with Jest and vue-jest


longnh301:
Code snippet:
```javascript
// menu.vue
export default {
  name: 'Menu',
  setup(props, context) {
    // logic here...
    return () => h('div', options, children)
  }
}

// test.js
const localVue = createLocalVue()
localVue.use(VueCompositionApi)

describe('Menu', () => {
  it('should render', () => {
    const wrapper = mount(Menu, {localVue})
    expect(wrapper.html()).toMatchSnapshot()
  })
})
```
which returns the following error
```
[Vue warn]: Failed to mount component: template or render function not defined. 
```
I'm currently using jest 24.9.0, @vue/test-utils and vue-jest 3.0.5. Other tests work fine with templated SFCs using composition API

sduduzog:
I just posted a question on stackoverflow for exacltly this. Here's a copy paste version of it
-----------

I am trying to test a jsx component by passing props to it and asserting that it renders them correctly. This fails though for a reason I cannot understand. Here's the component
```javascript
// components/toast.tsx

import { createComponent } from "@vue/composition-api";
import style from "./toast.module.scss";
export default createComponent({
  name: "Toast",
  props: {
    message: String
  },
  setup(props) {
    return () => (
      <div class={style.toast}>
        <p>{props.message}</p>
      </div>
    );
  }
});
```

Here's the snippet of the unit test that is failing:
```typescript
// tests/unit/components/toast.spec.ts

import Toast from "@/components/toast";
import { shallowMount } from "@vue/test-utils";

// ... a describe block

  it("should render props correctly", () => {
    const msg = "hello";
    const wrapper = shallowMount(Toast, {
      propsData: {
        message: msg
      }
    });
    expect(wrapper.isVueInstance()).toBeTruthy();
    expect(wrapper.text()).toMatch(msg);
  });
```

And the output from the failing test is here:
```
Expected substring: "hello"
    Received string:    "function (el) {
          var obj;·
          var args = [], len = arguments.length - 1;
          while ( len-- > 0 ) args[ len ] = arguments[ len + 1 ];
          if (shouldNotBeStubbed(el, stubs, modifiedComponents)) {
            return originalCreateElement.apply(void 0, [ el ].concat( args ))
          }·
          if (isConstructor(el) || isComponentOptions(el)) {
            if (stubAllComponents) {
[...]

PR #383, opened 2020-06-15
fix(setup): returning `render` function breaks when extending with an empty object

pikax:
fix https://github.com/vuejs/composition-api/issues/151

Fix the bug when returning render function: `Vue.extend(Comp).extend({}) - vue-test-utils`

The bug is caused because we try to merge the `function` with an empty object.

https://github.com/vuejs/vue-test-utils/pull/1512 can be reopen and merged after we release this
also fix https://github.com/vuejs/vue-test-utils/issues/1484

### DIFF
--- a/src/install.ts
+++ b/src/install.ts
@@ -9,6 +9,8 @@ import { VueConstructor } from 'vue'
  */
 function mergeData(from: AnyObject, to: AnyObject): Object {
   if (!from) return to
+  if (!to) return from
+
   let key: any
   let toVal: any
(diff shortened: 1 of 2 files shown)


lmiller1990:
Great! Does this mean that this will now work as expected without any changes to VTU?

pikax:
Yes, this should make it work without any changes to VTU

LeBenLeBen:
Thank you very much for taking the time to fix this! 🙇

Issue #384, opened 2020-06-16
markRaw does not prevent automatic ref unwrapping


vmihailenco:
`markRaw` should prevent automatic prop unwrapping in cases like

```js
      setup() {
        return { fooBar: markRaw({"ref": ref(123)}) }
      },
```

But it does not - see https://gist.github.com/vmihailenco/9ad24cf3492330fb31015fed3c88000b for full example.

Same code works fine with Vue 3.

PR #386, opened 2020-06-16
fix: not unwrapping `markRaw` objects

pikax:
fix https://github.com/vuejs/composition-api/issues/384

### DIFF
--- a/src/reactivity/unwrap.ts
+++ b/src/reactivity/unwrap.ts
@@ -1,6 +1,6 @@
 import { isRef } from './ref'
 import { proxy, isFunction, isPlainObject, isArray } from '../utils'
-import { isReactive } from './reactive'
+import { isReactive, isRaw } from './reactive'
 
 export function unwrapRefProxy(value: any, map = new WeakMap()) {
   if (map.has(value)) {
@@ -13,7 +13,8 @@ export function unwrapRefProxy(value: any, map = new WeakMap()) {
     isReactive(value) ||
     !isPlainObject(value) ||
     !Object.isExtensible(value) ||
-    isRef(value)
+    isRef(value) ||
+    isRaw(value)
   ) {
     return value
   }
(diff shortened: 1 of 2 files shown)


vmihailenco:
Thanks for fixing this.

Is there any way to undo unwrapping / get the target/underlying object? Something like `undo(props.foo)` which returns the target object which was not marked with `markRaw`.

pikax:
Because of the vue2 limitations, when  you change the `ref` it will change the underlying value

```ts
  it('test', ()=>{
    const a = {a:1 };
    const r = ref(a);

    r.value.a = 2

    expect(a.a).toBe(2)
  })
```

vmihailenco:
Let me re-phrase it. Automatic unwrapping practically changes object like `{foo: Ref<numer>}` into `{foo: number}` using a Proxy / get&set. So instead of `obj.foo.value` I should/can write `obj.foo`. Is there any way to get `{foo: Ref<numer>}` back from `{foo: number}`?

pikax:
Not currently possible. 

How do you do that in vue3?

vmihailenco:
>How do you do that in vue3?

I don't know - I thought you have an idea. Thanks for the help and sorry for the offtopic.

src/reactivity/unwrap.ts
```diff
@@ -28,8 +29,12 @@ export function unwrapRefProxy(value: any, map = new WeakMap()) {
 
   for (const k of Object.keys(value)) {
     const r = value[k]
+    // don't unwrap or do anything to raw
+    if (isRaw(r)) {
```
sapphi-red, line 33: I found that when `r` is `null` or `undefined`, it causes `TypeError: Cannot convert undefined or null to object`.

src/reactivity/unwrap.ts
```diff
@@ -29,8 +29,8 @@ export function unwrapRefProxy(value: any, map = new WeakMap()) {
 
   for (const k of Object.keys(value)) {
     const r = value[k]
-    // don't unwrap or do anything to raw
-    if (isRaw(r)) {
+    // don't process on falsy or raw
+    if (!r || isRaw(r)) {
```
antfu, line 33: Would it better to check falsy inside isRaw?

pikax:
```js
export function isRaw(obj: any): boolean {
  return !!obj && hasOwn(obj, RawIdentifierKey) && obj[RawIdentifierKey] === RawIdentifier;
}
```
Not sure if is will help, the idea is if is `isRaw` it shouldn't do anything to the object.

even if I add that condition on `isRaw` I still need to do `unwrap`, since we don't need to process `falsy` values

Issue #385, opened 2020-06-16
data() does not auto unwrap ref in template


vmihailenco:
E.g. this is properly unwrapped

```js
    setup() {
      return { fooBar: useFooBar() }
    },
```

But this is not unwrapped

```
    data() {
      return { fooBar: useFooBar() }
    },
```

PR #387, opened 2020-06-16
fix: unwrap refs returned by `data`

pikax:
fix https://github.com/vuejs/composition-api/issues/385

I have a few concerns here, because the unwrapping will create a new `object` everytime: 

```ts
const myObject = { a: 1}

const vm = new Vue({     data() {
        return {
          myObject,
        }
      },
    }).$mount()

vm.myObject === myObject; // this will always be false
```

Not sure how much of a deal breaker this might be, one way to do this is only `unwrap` if there's a `ref` anywhere in the object.

The reason we create a new object is because we don't want to mutate the original object the user is using in the `setup`, because that will lead to some unexpected bugs

### DIFF
--- a/src/setup.ts
+++ b/src/setup.ts
@@ -154,6 +154,22 @@ export function mixin(Vue: VueConstructor) {
       }
     }
 
+    const { data } = $options
+    // wrapper the data option, so we can invoke setup before data get resolved
+    $options.data = function wrappedData() {
+      if (setup) {
+        initSetup(vm, vm.$props)
+      }
+      const dataValue =
+        typeof data === 'function'
+          ? (data as (
+              this: ComponentInstance,
+              x: ComponentInstance
+            ) => object).call(vm, vm)
+          : data || {}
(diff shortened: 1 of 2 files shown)


antfu:
Thanks for make PR so fast as always!

Just tested out with vue-next, I am surprised to see that vue-next does handle that. (which makes `setup` and `data` almost the same to me). Since vue-next does that, I think it's good to have it in this plugin as well.

About the deep cloning, I think it's ok in most of the common cases. One thing I am worried about is the performance impact when passing a huge object. That would require users to use Object.freeze for them.

pikax:
My concerns is using the template to change the object values :/

antfu:
> template to change the object values

Sorry, I can't imagine a case right out of my mind. Can you show me an example about that?

pikax:
The case I was thinking  was this: https://github.com/vuejs/composition-api/pull/387/files#diff-de44c4f0babbf8fdd103f1756b4b7dd9R748 

But seems to work just fine

PR #399, opened 2020-06-22
BREAKING CHANGE: Change umd exported name to `VueCompositionAPI`

### LINKED ISSUE #396: RFC: Rename the UMD exported name
The current UMD name is `vueCompositionApi`.

I would propose changing to `VueCompositionAPI` to better align with our ecosystem (Vue, VueRoute, Vuex, etc).

antfu:
Close #396


antfu:
Currently missed APIs from @vue/runtime-core:

```
customRef
h
defineAsyncComponent
isProxy
isReadonly
isVNode
mergeProps
nextTick
onRenderTracked
onRenderTriggered
readonly
shallowReadonly
toHandlers
```

They should be either implemented or documented as not supporting.

Extra APIs

```js
createComponent // should be removed
createElement // should be renamed to h
onServerPrefetch // should be replaced with onRenderTracked and onRenderTriggered
set
```

PR #401, opened 2020-06-22
feat: export nextTick

antfu:
Align with vue-next

### DIFF
--- a/src/nextTick.ts
+++ b/src/nextTick.ts
@@ -1,13 +1,11 @@
 import Vue from 'vue'
-import { currentVM, currentVue } from './runtimeContext'
+import { currentVue } from './runtimeContext'
 
 type NextTick = Vue['$nextTick']
 
-export const nextTick: NextTick = function nextTick(...args: any) {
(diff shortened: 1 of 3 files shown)


src/nextTick.ts
```diff
@@ -0,0 +1,11 @@
+import Vue from 'vue'
+import { currentVue } from './runtimeContext'
+
+type NextTick = Vue['$nextTick']
+
+export const nextTick: NextTick = function nextTick(
+  this: ThisType<NextTick>,
+  ...args: Parameters<NextTick>
+) {
+  return currentVue?.nextTick.bind(this, args)
```
pikax, line 10: if called outside of a component setup/lifecycle `currentVue` won't have a value here, we should `Vue.nextTick` in that case

antfu: Yeah currentVue is Vue and it should always exist after installing, I guess currentVM might be what you are talking.

pikax: That's right, read as `currentVM` 😅

PR #405, opened 2020-06-23
fix(toRefs): do not warn when toRefs is called in a prop value

### LINKED ISSUE #393: Auto unwrapped template props are not considered reactive by toRefs
Following code

```js
toRefs(props.foo)
```

logs following error

```
toRefs() expects a reactive object but received a plain one
```

Even though `props.foo` *is* a reactive object because of template auto unwrapping and it makes sense to use `toRefs` on such props.

Same code works fine with vue-next.

pikax:
fix #393

### DIFF
--- a/test/setup.spec.js
+++ b/test/setup.spec.js
@@ -9,7 +9,6 @@ const {
   toRefs,
   markRaw,
 } = require('../src')
-const { nextTick } = require('process')
 
 describe('setup', () => {
   beforeEach(() => {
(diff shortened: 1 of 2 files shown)


test/setup.spec.js
```diff
@@ -9,6 +9,7 @@ const {
   toRefs,
   markRaw,
 } = require('../src')
+const { nextTick } = require('process')
```
antfu, line 12:
```suggestion
```

PR #406, opened 2020-06-23
fix(type): improve defineComponent type for option apis

### LINKED ISSUE #191: Access Data in render
https://github.com/vuejs/composition-api/blob/1672b6e214adb50a22e525c18cc9098953598024/src/component/component.ts#L79-L91

This is needed to get the `return` from `setup` to `render` as `this.data // RawBindings`.
```ts 
Omit<Vue2ComponentOptions<Vue, RawBindings>
```
If now there is a way to do it right, I would like to know it.

### LINKED ISSUE #338: TypeScript with composition-api: in methods, "this" is Vue, not current instance
### Version
2.6.11

### Reproduction link
[https://codepen.io/robertgres/pen/XWmozGV](https://codepen.io/robertgres/pen/XWmozGV)

### Steps to reproduce
```
import * as vue from '@vue/composition-api'

vue.defineComponent({
  methods: {
    foo() {
       this.baz() // Error
    },
    baz() {}
  },
  mounted() {
     this.foo() // fine
  }
})
```

### What is expected?
No TypeScript errors

### What is actually happening?
TS2339: Property 'baz' does not exist on type 'Vue'.

---
Seems like typings don't take into accout current instance

This fixes given problem (in file types/options.d.ts):

`type DefaultMethods<V, S extends V> =  { [key: string]: (this: S, ...args: any[]) => any };`

instead of:

`type DefaultMethods<V> =  { [key: string]: (this: V, ...args: any[]) => any };`

antfu:
Resolve #338, Resolve #191

This should fix the most common use cases for type inference in `defineComponent`. Type inference of `mixins` and `emit` is not implemented in this PR.


antfu:
@pikax Can you help review a bit? Sorry for being kinda huge.

I took a lot of code from vue-next as references. I also copied the `defineComponent.test-d.ts` from it, except for the `mixins` and `emit` parts. Tests are passing, so I think it should work fine. Just pinning you to be sure. Thank you. 

This was hard and I have to say I learned Typescript better 😅

Aaron-Pool:
Thanks so much for getting a quick PR for this going, @antfu! It's super appreciated 🙇

antfu:
Force pushed against `master` as this is not a breaking change. We can release the fix before 1.0.

PR #409, opened 2020-06-25
fix(unwrapRef): copy __ob__ and make toRaw work in props

### LINKED ISSUE #392: toRaw does not undo props auto unwrapping
composition-api v0.6.5 toRaw does not unwrap reactive:

```js
toRaw(reactive({ "ref": ref(1) }))
```

In vue-next same code works as expected.

antfu:
fix #392

### DIFF
--- a/src/reactivity/unwrap.ts
+++ b/src/reactivity/unwrap.ts
@@ -27,6 +27,11 @@ export function unwrapRefProxy(value: any, map = new WeakMap()) {
     (s) => (obj[s] = (value as any)[s])
   )
 
+  // copy __ob__
+  if (value.__ob__) {
+    Object.defineProperty(obj, '__ob__', value.__ob__)
+  }
+
   for (const k of Object.keys(value)) {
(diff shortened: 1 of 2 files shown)


src/reactivity/unwrap.ts
```diff
@@ -27,6 +27,11 @@ export function unwrapRefProxy(value: any, map = new WeakMap()) {
     (s) => (obj[s] = (value as any)[s])
   )
 
+  // copy __ob__
+  if (value.__ob__) {
```
pikax, line 31: shouldn't that be part of the `Object.keys`? if so can we just add it there when the key is `__ob__`?

antfu: [It's defined by `defineProperty`](https://github.com/vuejs/vue/blob/0b31647c41b0c916d40b962723b1de06672d0e01/src/core/observer/index.js#L46) which is not enumerable by `Object.keys`

PR #413, opened 2020-06-28
BREAKING CHANGE: match dist file naming with vue 2

antfu:
match with vue 2 naming style.

Now the dist file would look this:

image

cjs will import dev/prod version based on current env just acting the same behavior as vue. 

This should not affect regular users as the indexes in package.json changed as well.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,0 +1,7 @@
+'use strict'
+
+if (process.env.NODE_ENV === 'production') {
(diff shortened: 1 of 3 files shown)


posva:
doesn't it make more sense to keep the naming consistent with Vue 2 dist files instead?

antfu:
Updated.

PR #420, opened 2020-07-03
feat: add useCSSModule api

### LINKED ISSUE #21: Cannot work with CSS Modules
```Vue
<script>
  export default {
    props: {
      themeName: String
    },

    computed: {
      $theme () {
        return this[this.themeName]
      }
    }
  }
</script>

<style module="$a">
  .wrapper {
    background: #000;
  }
</style>

<style module="$b">
  .wrapper {
    background: #fff;
  }
</style>
```

`setup (props, context)` cannot work with CSS Modules, because `context` does not has property `$a` or `$b`.

yoyo930021:
Fixed #21


antfu:
Thanks for the PR! 

But sorry I don't think it belongs to the Composition API. And since it only relies on the public API `getCurrentInstance`, I think you can create a library of it and publish to npm for people to use. :)

yoyo930021:
It is built-in in `vue-next`.
[link](https://github.com/vuejs/vue-next/blob/master/packages/runtime-core/src/helpers/useCssModule.ts)
So it will help us migrate to vue 3.

yoyo930021:
> Can you add some tests?

Hmm...
I cannot find related tests in `vue-next`.

We maybe need `vue-loader` and compile SFC file for the test.
I will study when free time.

src/apis/useCssModule.ts
```diff
@@ -0,0 +1,23 @@
...
+  }
+
+  const mod = (instance as Record<string, string>)[name]
+  if (!mod) {
+    __DEV__ &&
+      warn(`Current instance does not have CSS module named "${name}".`)
+    return EMPTY_OBJ
+  }
+
+  return (mod as unknown) as Record<string, string>
```
antfu, line 22: Assigning `instance` to `Record<string, string>` makes `mod` a string, which forces you to do the type reset. Not a big deal though.

```suggestion
  const mod = (instance as any)[name]
  if (!mod) {
    __DEV__ &&
      warn(`Current instance does not have CSS module named "${name}".`)
    return EMPTY_OBJ
  }

  return mod as Record<string, string>
```

yoyo930021: It's my mistake.

yoyo930021:
> LGTM, thanks.
> 
> I don't think tests are must-have here since the only thing it does is to expose an attribute. It mainly depends on the bundling environments, setting up tests for this particular case costs a bit too much but little to gain. 

I agree with you,
So I didn't add any tests at the first time.
😆

dodas:
Is there some documentation, explaining how this should be used?
Thanks.

antfu:
@dodas https://vue-loader.vuejs.org/guide/css-modules.html

dodas:
Ahh, so `useCssModule()` just returns what you currently have under `this.$style`, right?
Thanks.

antfu:
Yes

Issue #437, opened 2020-07-11
Where is the cloneVNode function?


ddenev:
Evan You says we can use the cloneVNode function - https://github.com/vuejs/vue-next/issues/815#issuecomment-598255232 - but I cannot find it in the composition API.

Tried `import { cloneVNode } from 'vue'` and `import { cloneVNode } from '@vue/composition-api'` but none works.

I'm using Vue 2 with the `@vue/composition-api` package so what is the proper way to import the `cloneVNode` function?

Thank you.

pikax:
That's quite low-level API, I've created a draft PR, but the implementation is hacky since vue 2.x doesn't expose this internals in a good way. 

Even if it gets ported there will be some limitations.

pikax:
`cloneVNode` and other low level APIs are not part of the composition-api, so they won't be supported here.

Also the vdom of v2 is different than v3 making the port of this api not ideal.

PR #438, opened 2020-07-13
feat: vNode helpers parity

pikax:
fix https://github.com/vuejs/composition-api/issues/437

Adds missing vnode functions.

Implementation is a bit hacky since vue 2.x doesn't expose vnode and other helpers


pikax:
Closing because this is not really composition api and the vdom of v2 is different than v3

k644606347:
excuse me, what do I need to do if I want to use cloneVNode? My project cannot be upgraded vue3;

copy cloneVNode to my project?

PR #442, opened 2020-07-14
fix(IE11): replace `startsWith` to be IE11 compatible

pikax:
IE11 doesn't support `startsWith`, replaced it with array check.

### DIFF
--- a/src/install.ts
+++ b/src/install.ts
@@ -55,7 +55,7 @@ export function install(Vue: VueConstructor) {
   }
 
   if (__DEV__) {
-    if (Vue.version[0] !== '2' && Vue.version[1] !== '.') {
+    if (Vue.version[0] !== '2' || Vue.version[1] !== '.') {
       assert(false, `only works with Vue 2, v${Vue.version} found.`)
     }
   }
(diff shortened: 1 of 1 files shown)


src/install.ts
```diff
@@ -55,7 +55,7 @@ export function install(Vue: VueConstructor) {
   }
 
   if (__DEV__) {
-    if (!Vue.version.startsWith('2.')) {
+    if (Vue.version[0] !== '2') {
```
antfu, line 58: It's probably ok as we would not expect to see Vue 20 or something, but as it costs almost nothing. I would suggest having a check at `[1] === '.'` to be more careful. 😅

pikax: Thought about it too :D

src/install.ts
```diff
@@ -55,7 +55,7 @@ export function install(Vue: VueConstructor) {
   }
 
   if (__DEV__) {
-    if (!Vue.version.startsWith('2.')) {
+    if (Vue.version[0] !== '2' && Vue.version[1] !== '.') {
```
jacekkarczmarczyk, line 58:
```suggestion
    if (Vue.version[0] !== '2' || Vue.version[1] !== '.') {
```

pikax: Thank you 😅

Issue #448, opened 2020-07-23
Infinite update loop in 0.6.2 - 1.0.0-beta.6


drafu:
The bug triggering "You may have an infinite update loop in a component render function" happens in unexpected places on unexpected reasons.

It did not happen with @vue/composition-api up until 0.6.1 and started happening in 0.6.2 and still persists.

I have recreated a minimal app with this problem
https://github.com/drafu/vue-composition-infinite-loop

**This bug strangely disappears if I:**
Remove A.vue component from App.vue
**OR**
Remove pattern prop from inside of A.vue component
**OR**
Remove div v-for iteration in App.vue
**OR**
Remove B.vue component from App.vue
**OR**
Remove emit from onMounted from B.vue

Which is highly strange all things considered, like, how is A.vue component, which is completely static, triggers infinite loop.

This bug is happening in some other places in my projects, but I cannot trace any consistent pattern.

kahirokunn:
Excuse me, it looks like a test has been added, is @vue/composition-api@latest already solve this problem? :eyes:

PR #449, opened 2020-07-23
fix: Infinite loop on a edge case when props value instance is created each render

pikax:
fix https://github.com/vuejs/composition-api/issues/448

I still don't know exactly why that is happening, from my findings is caused by the `props.pattern` is assigned a new value each render, I think this is an edge case. 

The props are mark as `reactive` only because `toRefs` will be complaining if they aren't, so adding a new propsReactive and checking that on the `toRefs` makes all test passing

### DIFF
--- a/src/mixin.ts
+++ b/src/mixin.ts
@@ -21,6 +21,7 @@ import {
   resolveScopedSlots,
   asVmProperty,
 } from './utils/instance'
+import { markPropsReactive } from './reactivity/reactive'
 
 export function mixin(Vue: VueConstructor) {
   Vue.mixin({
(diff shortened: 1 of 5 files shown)


antfu:
Thanks for looking into it but I doubt it's a good solution. My guess is that props become reactive while setup will try to unwrap it by creating a new object. When we doing updating, the reference checking failed, which caused the infinite loop.

On my trivial trying, adding a props reference check to the following line the test passed

```ts
          if (isReactive(bindingValue) && bindingValue != props) { // <--
            bindingValue = ref(bindingValue)
          } else {
```

https://github.com/vuejs/composition-api/blob/16c6a60ce61175051c8ded8b3b84acd1dc08b3bb/src/mixin.ts#L112-L118

kahirokunn:
Is there any progress?

pikax:
@kahirokunn I'm still looking for a better solution.

The error occurs in this edge case, if you know anymore cases please let me know

PR #456, opened 2020-08-04
feat: `proxyRefs` method and `ShallowUnwrapRefs` type

pikax:
https://github.com/vuejs/vue-next/pull/1682


test/setup.spec.js
```diff
@@ -177,54 +177,54 @@ describe('setup', () => {
...
-    expect(vm.$el.textContent).toBe('31')
-
-    bar.value = 4
-    baz.value = 2
-
-    await vm.$nextTick()
-    expect(warn).not.toHaveBeenCalled()
-    expect(vm.$el.textContent).toBe('42')
-  })
+  // it('not warn doing toRef on props', async () => {
```
pikax, line 180: this is caused because the props are marked as `raw`

Open for suggestions

antfu: Guess we can mark the props to shallow reactive and change the test to only tests the top-level `toRefs`, just like you mentioned in the limitation part?

pikax: Props are already reactive, only the nested ref are not
https://github.com/vuejs/composition-api/blob/030d02ca0f67782629dc602f2f441a8712f78bb7/test/setup.spec.js#L336-L371

README.md
```diff
@@ -392,6 +392,17 @@ app2.component('Bar', Bar) // equivalent to Vue.use('Bar', Bar)
 
 </details>
 
+### props
+<details>
+<summary>
+⚠️ <code>toRefs(props.foo.bar)</code> will incorrectly warn when acessing nested levels of props.
+⚠️ <code>isReactive(props.foo.bar)</code> will return false.
+</summary>
+TODO add description
```
antfu, line 401:
```suggestion
\`\`\`
defineComponent({
  setup(props) {
    const { bar } = toRefs(props.foo) // won't work
    
    // use this instead 
    const { foo } = toRefs(props)
    const a = foo.value.bar
  }
})
\`\`\`
```

src/mixin.ts
```diff
@@ -105,6 +99,8 @@ export function mixin(Vue: VueConstructor) {
       const bindingObj = binding
       vmStateManager.set(vm, 'rawBindings', binding)
 
+      // binding = proxyRefs(binding);
+
```
antfu, line 103:
```suggestion
      // binding = proxyRefs(binding);

```

Guess it's temporary?

test/setup.spec.js
antfu: I think it's fine by leaving a note in the limitation for this.

src/reactivity/ref.ts
```diff
@@ -185,3 +187,33 @@ export function triggerRef(value: any) {
 
   value.value = value.value
 }
+
+export function proxyRefs<T extends object>(
+  objectWithRefs: T
+): ShallowUnwrapRef<T> {
+  if (isReactive(objectWithRefs)) {
+    //@ts-ignore
```
antfu, line 195: Maybe we should use `as` instead of `@ts-ignore`?

PR #468, opened 2020-08-07
fix(watch): watch will trigger when added new keys using `set`

### LINKED ISSUE #467: watch() with reactive() doesn't re-run when new properties are added to the reactive object
https://codesandbox.io/s/dry-leftpad-xrbtl

pikax:
fix #467

### DIFF
--- a/README.md
+++ b/README.md
@@ -176,6 +176,27 @@ a.list[1].count === 1 // true
 
 </details>
 
+<details1>
+<summary>
+⚠️ adding new reactive properties `Vue.set`
+</summary>
+
+> Part of the limitation of [Vue 2.x reactivity](https://vuejs.org/v2/guide/reactivity.html#For-Objects) is to track new keys on an `object` you need to call `Vue.set`.
+
+```ts
+import { reactive, set } from '@vue/composition-api'
+
+const a = reactive({
+  foo: 1
+})
+
+// add new reactive key
+set(a, 'bar', 1)
+```
+
+
+</details>
+
 ### Template Refs
 
 <details>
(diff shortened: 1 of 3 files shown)


antfu:
If I have recalled right, Vue 3 doesn't offer the `set` api. As the reactivity systems underneath are different, I think the `set` api is useful in some situations with Vue 2. Should we add a note in the readme to mention this difference?

pikax:
`set` is only to circumvent the limitation on adding a new `key` limitation on `vue2`. In `vue3` there's no such limitation when using proxys.

We should add it to the readme.

README.md
```diff
@@ -174,6 +174,27 @@ a.list.push(
...
 ```
 
+</details>
+
+<details1>
+<summary>
+⚠️ adding new reactive properties `Vue.set`
+</summary>
+
+> Part of the limitation of [Vue 2.x reactivity](https://vuejs.org/v2/guide/reactivity.html#For-Objects) is to track new keys on an `object` you need to call `Vue.set`.
```
antfu, line 184:
```suggestion
<summary>
⚠️ `set` workaround for adding new reactive properties
</summary>

> ⚠️ Warning: `set` does NOT exist in Vue 3. We provide it as a workaround here, due to the limitation of [Vue 2.x reactivity system](https://vuejs.org/v2/guide/reactivity.html#For-Objects). In Vue 2, you will need to call `set` to track new keys on an `object`(similar to `Vue.set` but for `reactive objects` created by the Composition API). In Vue 3, you can just assign them like normal objects.
```

README.md
```diff
@@ -174,6 +174,27 @@ a.list.push(
 a.list[1].count === 1 // true
 ```
 
+</details>
+
+<details1>
```
antfu, line 179:
```suggestion
<details>
```
😆

PR #476, opened 2020-08-13
docs: added minzipped size badge

privatenumber:
Bundle size is often one of the first questions asked when deciding to add a dependency.

Added a bundle-size badge to advertise the small size.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,6 +4,9 @@ Vue 2 plugin for **Composition API**
 
 [![npm](https://img.shields.io/npm/v/@vue/composition-api)](https://www.npmjs.com/package/@vue/composition-api)
 [![GitHub Workflow Status](https://img.shields.io/github/workflow/status/vuejs/composition-api/Build%20&%20Test)](https://github.com/vuejs/composition-api/actions?query=workflow%3A%22Build+%26+Test%22)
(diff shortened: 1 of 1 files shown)


posva:
Thanks but bundlephobia is not accurate when dealing with libraries that can be tree shaken like the composition api or Vue 3

More realistic tests can be done with size-limit or by adapting its approach

privatenumber:
Bundlephobia has a new Exports Analysis section that details the size of each export: 

https://bundlephobia.com/result?p=@vue/composition-api


(Vue 3 doesn't seem to have `sideeffects: false` for [Bundlephobia to analyze](https://bundlephobia.com/result?p=vue@3.0.0-rc.5))

antfu:
@vue/composition-api has a global mixin which makes the exports almost identical which comes to tree-shaking. I think it's fine to have it as ppl can also click in an check the details.

pikax:
> Bundlephobia has a new Exports Analysis section that details the size of each export:
> 
> https://bundlephobia.com/result?p=@vue/composition-api
> 
> (Vue 3 doesn't seem to have `sideeffects: false` for [Bundlephobia to analyze](https://bundlephobia.com/result?p=vue@3.0.0-rc.5))

Effect free you need to use [@vue/runtime-core](https://bundlephobia.com/result?p=@vue/runtime-core@3.0.0-rc.5)

As for composition-api, this is what posva was referring to:
image

As you can notice all the functions have the same size (which is the total size of the lib)

PR #477, opened 2020-08-14
fix(watch): using deep:true in SSR context

### LINKED ISSUE #473: Error in data(): "TypeError: Cannot read property 'dep' of undefined" in beta9
A project that had no errors in beta8 is experiencing the following errors in beta9.
I am not using the prop called `dep` (it also same with my modules)
and the problem was solved when I went back to beta8.

> Terminal

스크린샷 2020-08-13 오후 4 34 10

> package.json
```
"@vue/composition-api": "^1.0.0-beta.9"
```

> client/pages/index.vue

스크린샷 2020-08-13 오후 4 38 46

> .bind() (vue-state-store/src/store.ts)
스크린샷 2020-08-13 오후 4 40 31

Maybe it can be possible the setting `deep` is used in the composition api as a `dep`,
and some changes makes the problem and that is might be problem...?

pikax:
fix #473 

I don't know what is causing this to happen, so I can't test this.

This should never happen because the `watcher.value` is still reactive but doesn't have `__ob__`

### DIFF
--- a/src/apis/watch.ts
+++ b/src/apis/watch.ts
@@ -321,7 +321,7 @@ function createWatcher(
 
   // if the return value is reactive and deep:true
   // watch for changes, this might happen when new key is added
-  if (isReactive(watcher.value) && deep) {
+  if (isReactive(watcher.value) && watcher.value.__ob__?.dep && deep) {
     watcher.value.__ob__.dep.addSub({
       update() {
         // this will force the source to be revaluated and the callback


pikax:
Found the issue!

This is caused by SSR rendering.

Minimal example with nuxt:  https://codesandbox.io/s/determined-ganguly-vfdei?file=/pages/index.vue

PR #481, opened 2020-08-18
Adjusted the inject function code to be flat like the provide function

kahirokunn:


src/apis/inject.ts
```diff
@@ -48,17 +48,18 @@ export function inject(
...
+
+  const val = resolveInject(key, vm)
+  if (val === NOT_FOUND) {
+    if (defaultValue === undefined && __DEV__) {
+      warn(`Injection "${String(key)}" not found`, vm)
+    }
+    return defaultValue
   }
+
+  return val
```
jacekkarczmarczyk, line 64: how about even more flat?

```suggestion
  if (val !== NOT_FOUND) {
    return val;
  }
  
  if (defaultValue === undefined && __DEV__) {
    warn(`Injection "${String(key)}" not found`, vm)
  }

  return defaultValue
```

kahirokunn: That's great!

PR #497, opened 2020-09-02
feat: allow plugin to be installed in localVue

### LINKED ISSUE #492: Bug: Cannot scope install Composition API in the `localVue` by `@vue/test-utils`
Hi,

I'm trying to test my component that is using `@vue/composition-api`, and I'm using `@vue/test-utils` as the test library.

Based on `@vue/test-utils`, it provides `createLocalVue` to copy the original Vue instance, so we don't pollute the original `Vue` instance if we need to install specific plugin.

https://vue-test-utils.vuejs.org/api/options.html#localvue

But when I tried to install the plugin to different copies from `localVue` instance, it still prohibits me from installing the composition API plugin to another instance.

Code example:

```
import CompositionApi from '@vue/composition-api';
import { createLocalVue } from '@vue/test-utils';

// create first copy
const localVueOne = createLocalVue();
localVueOne.use(CompositionApi);  // this one is fine

// create second copy
const localVueTwo = createLocalVue();
localVueOne.use(CompositionApi);  // this one triggers me warning that the composition API is already installed
```

I suspect that it's because there's a "flag" to check if the plugin is already installed, without checking if it's installing from the same instance or not.

https://github.com/vuejs/composition-api/blob/026a78aab73df4f1c601a085dc8aa2cbe9e8bcb4/src/runtimeContext.ts#L10

My expectation was that the copy should still able to install the plugin again, just like other plugin (Vuex/VueRouter/etc).

Please confirm if it's the correct behavior or not. Thanks!

pikax:
fix #492

Need a better error message when replacing the VueConstructor,

I haven't tested but I don't think the plugin will play nice with multiple vue instances running at the same time.

### DIFF
--- a/src/install.ts
+++ b/src/install.ts
@@ -1,6 +1,6 @@
 import type { VueConstructor } from 'vue'
 import { AnyObject } from './types/basic'
-import { hasSymbol, hasOwn, isPlainObject, assert } from './utils'
+import { hasSymbol, hasOwn, isPlainObject, assert, warn } from './utils'
 import { isRef } from './reactivity'
 import { setVueConstructor, isVueRegistered } from './runtimeContext'
 import { mixin } from './mixin'
@@ -51,8 +51,12 @@ export function install(Vue: VueConstructor) {
   }
 
   if (__DEV__) {
-    if (Vue.version && (Vue.version[0] !== '2' || Vue.version[1] !== '.')) {
-      assert(false, `only works with Vue 2, v${Vue.version} found.`)
+    if (Vue.version) {
+      if (Vue.version[0] !== '2' || Vue.version[1] !== '.') {
+        assert(false, `only works with Vue 2, v${Vue.version} found.`)
+      }
+    } else {
+      warn('Vue version not found')
     }
   }
 
--- a/src/install.ts
+++ b/src/install.ts
@@ -2,11 +2,7 @@ import type { VueConstructor } from 'vue'
 import { AnyObject } from './types/basic'
 import { hasSymbol, hasOwn, isPlainObject, assert } from './utils'
 import { isRef } from './reactivity'
-import {
-  setVueConstructor,
-  isVueRegistered,
-  isPluginInstalled,
-} from './runtimeContext'
+import { setVueConstructor, isVueRegistered } from './runtimeContext'
 import { mixin } from './mixin'
 
 /**
@@ -44,7 +40,7 @@ function mergeData(from: AnyObject, to: AnyObject): Object {
 }
 
 export function install(Vue: VueConstructor) {
-  if (isPluginInstalled() || isVueRegistered(Vue)) {
+  if (isVueRegistered(Vue)) {
     if (__DEV__) {
       assert(
         false,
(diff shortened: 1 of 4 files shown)


src/runtimeContext.ts
```diff
@@ -27,6 +27,12 @@ export function getVueConstructor(): VueConstructor {
 }
 
 export function setVueConstructor(Vue: VueConstructor) {
+  if (__DEV__ && vueConstructor) {
+    // don't warn in jest
+    if ('jest' in global) {
```
antfu, line 32: Is this too special? I think it might need to be handle on the user side.

src/install.ts
```diff
@@ -55,7 +51,7 @@ export function install(Vue: VueConstructor) {
   }
 
   if (__DEV__) {
-    if (Vue.version[0] !== '2' || Vue.version[1] !== '.') {
+    if (Vue.version && (Vue.version[0] !== '2' || Vue.version[1] !== '.')) {
```
antfu, line 54: This is nice to have, however, should we throw another warning when Vue.version doesn't exist?

src/runtimeContext.ts
pikax: I feel it would be annoying you getting the warning when testing, because you will always have the warning if you use `localVue`

src/install.ts
pikax: When you do `Vue.extend` it doesn't come with the version :/ 

Currently it throws an exception because version is `undefined`

src/runtimeContext.ts
antfu: Hmm, then how about `ava` `mocha`, etc? I am not very sure about this.

pikax: me neither, we have two choices, warn every time or not warn at all

src/install.ts
antfu: ok. Should we validate whether the `Vue` passed is a valid Vue constructor or we just let it go? Is there any way to do that?

src/runtimeContext.ts
antfu: I would stay just warn every time. And let users who wrote the tests to handle it. (didn't try it, but I guess ava runs in isolated env parallelly, which might not have this issue). Anyway, I feel it's better to warn as ppl might need to be sure what they are doing.

src/install.ts
pikax: Probably better just warn if `Vue.version` is not present, what do you think?

antfu: Good to me :)

test/use.spec.ts
```diff
@@ -0,0 +1,38 @@
+import CompositionApi from '../src'
+import { createLocalVue } from './helpers/create-local-vue'
+import { mockWarn } from './helpers'
+
+describe('use', () => {
+  mockWarn(true)
+
+  it('should allow install in multiple vue', () => {
+    // @ts-ignore
+    global.jest = undefined
```
antfu, line 10:
```suggestion
```

test/use.spec.ts
```diff
@@ -0,0 +1,38 @@
...
+    const localVueOne = createLocalVue()
+    localVueOne.use(CompositionApi)
+
+    expect(() => {
+      // vue prevents the same plugin of being installed, this will create a new plugin instance
+      localVueOne.use({
+        install() {
+          //@ts-ignore
+          CompositionApi.install(...arguments)
+        },
```
antfu, line 28:
```suggestion
        install(...args: any[]) {
          CompositionApi.install(...args)
        },
```

How about this?

antfu (approved):

PR #502, opened 2020-09-04
feat: improve reactive checks

### LINKED ISSUE #219: "reactive" doesn't work well with Array
The `reactive` can't observe a `Array`, my code like this:
```javascript
<template>
  <div>
    <button @click="handleAdd">Add count</button>
  </div>
</template>
<script>
import Vue from "vue";
import { setup, reactive, ref, watch } from "@vue/composition-api";

const observableArray = Vue.observable([{ value: 1 }]);

export default {
  setup() {
    const reactiveArray = reactive([{ value: 1 }]);
    const reactiveObject = reactive({ value: 1 });
    const refArray = ref([{ value: 1 }]);
    const reactiveDeepArray = reactive({parent: [{ value: 1 }]})

    watch(
      () => reactiveArray[0].value,
      val => {
        // I think this should be execute when I click the button every time,
        // but it doesn't
        console.log("reactiveArray changed", val);
      }
    );

    watch(
      () => reactiveObject.value,
      val => {
        // This will execute when I click the button every time
        console.log("reactiveObject changed", val);
      }
    );

    watch(
      () => refArray.value[0].value,
      val => {
        // This will execute when I click the button every time
        console.log("refArray changed", val);
      }
    );

    watch(
      () => observableArray[0].value,
      val => {
        // This will execute when I click the button every time
        console.log("observableArray changed", val);
      }
    );
    
    watch(
      () => reactiveDeepArray.parent[0].value,
      val => {
        // This will execute when I click the button every time
        console.log("reactiveDeepArray changed", val);
      }
    );

    function handleAdd() {
      reactiveArray[0].value++
      reactiveObject.value++
      refArray.value[0].value++
      observableArray[0].value++
      reactiveDeepArray.parent[0].value++
    }

    return {
      handleAdd
    };
  }
};
</script>

```
So, when I pass a `Array` to `reactive`, it won't be observed. But if the Array is wrapped by a `object`, it can be observed. I think it's a bug. If it's a feature, why?

pikax:
Instead of changing `reactive` object we just use a weak set to keep track on what objects are reactive

This allows to have `arrays` as `reactive`, bringing it closer to vue 3

Still need to:
- Extensive testing

Close #501, close #219

### DIFF
--- a/src/reactivity/reactive.ts
+++ b/src/reactivity/reactive.ts
@@ -2,31 +2,20 @@ import { AnyObject } from '../types/basic'
 import { getVueConstructor } from '../runtimeContext'
 import { isPlainObject, def, warn, isFunction } from '../utils'
 import { isComponentInstance, defineComponentInstance } from '../utils/helper'
-import { ReadonlyIdentifierKey, RefKey } from '../utils/symbols'
+import { RefKey } from '../utils/symbols'
 import { isRef, UnwrapRef } from './ref'
 import { rawSet, readonlySet, reactiveSet } from '../utils/sets'
 
 export function isRaw(obj: any): boolean {
   return rawSet.has(obj)
-  // return (
-  //   hasOwn(obj, RawIdentifierKey) && obj[RawIdentifierKey] === RawIdentifier
-  // )
 }
 
 export function isReadonly(obj: any): boolean {
   return readonlySet.has(obj)
-  // return hasOwn(obj, ReadonlyIdentifierKey) && obj[ReadonlyIdentifierKey]
 }
 
 export function isReactive(obj: any): boolean {
   return reactiveSet.has(obj)
-
-  // return (
-  //   isObject(obj) &&
-  //   Object.isExtensible(obj) &&
-  //   hasOwn(obj, ReactiveIdentifierKey) &&
-  //   obj[ReactiveIdentifierKey] === ReactiveIdentifier
-  // )
 }
 
 /**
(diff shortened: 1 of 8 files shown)


test/v3/reactivity/reactive.spec.ts
```diff
@@ -57,8 +57,8 @@ describe('reactivity/reactive', () => {
     }
     const observed = reactive(original)
     expect(isReactive(observed.nested)).toBe(true)
-    // expect(isReactive(observed.array)).toBe(true); //not supported by vue2
-    // expect(isReactive(observed.array[0])).toBe(true); //not supported by vue2
+    expect(isReactive(observed.array)).toBe(true) //not supported by vue2
+    expect(isReactive(observed.array[0])).toBe(true) //not supported by vue2
```
antfu, line 61:
```suggestion
    expect(isReactive(observed.array)).toBe(true)
    expect(isReactive(observed.array[0])).toBe(true)
```

antfu (approved):

src/reactivity/reactive.ts
```diff
@@ -203,29 +176,47 @@ export function shallowReactive<T extends object = any>(obj: T): T {
...
+      isFunction(target.__ob__.dep.addSub)
+    ) {
+      // @ts-ignore
+      target.__ob__.dep.addSub({
+        update() {
+          target.forEach((x) => markReactive(x))
+        },
+      })
+    }
+    // TODO way to track new array items
```
antfu, line 215: Should we hijack the array methods or just leave a caveat in the readme?

pikax: Is quite interesting idea, just thinking we might need to listen to other "events" to keep it consistent, for example, if you assign an array to an object, are we hijacking that array?

antfu: What do you mean "events"? If I recall correctly, Vue 2 hijacked the array's push/unshift, etc to achieve the reactivity updates.

pikax: when you add a new property to an object or when you change the property of the obj to an array

antfu: Oh, I see, I got what you mean. And yeah, if we handle the hijacking here and wrap the value with `markReactive` for newly added properties, they should be well covered I think.

PR #509, opened 2020-09-11
feat(reactivity): add support for map and set

pikax:
Similar implementation as `vue-next`

### DIFF
--- a/README.md
+++ b/README.md
@@ -154,33 +154,6 @@ b.list[0].count.value === 0 // true
 
 <details>
 <summary>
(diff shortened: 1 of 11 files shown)


Aaron-Pool:
@pikax do you have any kind of time table of when this might be merged? Or is this PR abandoned?

pikax:
@Aaron-Pool yeah pretty much, when doing this PR I noticed that more access to vue reactivity internals were needed, but there's no way to do that on vue 2.6, so I'm holding this for the port of [composition-api to vuejs core](https://github.com/vuejs/vue/pull/12000)

PR #511, opened 2020-09-12
fix(sets): check for window to avoid SSR errors

manniL:
Hey 👋🏻

The new version unfortunately breaks when using the composition API on server-side because of a missing check for `typeof window !== 'undefined'`.

image

This PR should add the check and avoid SSR errors 😋


antfu:
Fix released in `1.0.0-beta.13`, thanks for catching it!

PR #512, opened 2020-09-12
feat: making all vue2 reactive object be valid isReactive

### LINKED ISSUE #517: App breaks after upgrade from `1.0.0-beta.11` to `1.0.0-beta.12` and later
We have a web app in this [GitHub repo](https://github.com/DarkLite1/hip) that works perfectly fine with `@vue/composition-api` version `1.0.0-beta.11` but not after upgrading to `1.0.0-beta.12` or `13`. Navigation breaks and pages don't get loaded:

Given that the only thing changed is the ugprade of the vue composition api I'm wondering what changed or what we dan do to make it work again? It seems hard to troubleshoot. We think it might be related to [calling the graphql backend](https://github.com/DarkLite1/hip/blob/master/src/pages/Index.vue) as at that page the data is no longer loaded/fetched and the web-page blocks and keeps on showing "All accounts: loading" which means it's not getting data from the backend.

In a normal flow a promise is triggered after authentication ([handleResponse](https://github.com/DarkLite1/hip/blob/master/src/services/auth/authService.ts)) that sets the Vue reactive property ([setAccount](https://github.com/DarkLite1/hip/blob/master/src/store/authStore.ts)) for authentication and the home page is loaded and a graphql query is triggered. But now, after a the upgrade it's stuck here and navigating away from this wrong page seems to be not possible.

image

Any help would be appreciated.

image

image

pikax:
fix #517 

Simplifying the logic of `isReactive` now every valid reactive v2 object will be a valid isReactive object.

The only issue with this is `isReactive(obj) === true` won't mean auto-unwrapping 

```ts
const obj = Vue.observable({ a: ref(1) })
const reactiveObj = reactive({a: ref(1) })

isReactive(obj) // true

isReactive(reactiveObj) // true


obj.a // {value:1}
reactiveObj.a // 1

```

### DIFF
--- a/src/mixin.ts
+++ b/src/mixin.ts
@@ -5,7 +5,7 @@ import {
   SetupFunction,
   Data,
 } from './component'
-import { isRef, isReactive, markRaw, markReactive, toRefs } from './reactivity'
+import { isRef, isReactive, markRaw, toRefs } from './reactivity'
 import { isPlainObject, assert, proxy, warn, isFunction } from './utils'
 import { ref } from './apis'
 import vmStateManager from './utils/vmStateManager'
@@ -74,7 +74,8 @@ export function mixin(Vue: VueConstructor) {
     const ctx = createSetupContext(vm)
 
     // mark props
-    markReactive(props)
+    // markReactive(props)
+    props.__ob__ = true // fake reactive
 
     // resolve scopedSlots and slots to functions
     resolveScopedSlots(vm, ctx.slots)
--- a/src/mixin.ts
+++ b/src/mixin.ts
@@ -5,8 +5,15 @@ import {
   SetupFunction,
   Data,
 } from './component'
-import { isRef, isReactive, markRaw, toRefs } from './reactivity'
-import { isPlainObject, assert, proxy, warn, isFunction } from './utils'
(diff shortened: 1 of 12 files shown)


src/mixin.ts
```diff
@@ -73,14 +80,15 @@ export function mixin(Vue: VueConstructor) {
     const setup = vm.$options.setup!
     const ctx = createSetupContext(vm)
 
-    // mark props
-    markReactive(props)
+    // fake reactive for `toRefs(props)`
+    props.__props_reactive__ = true
```
antfu, line 84: maybe it's better to use `def` here?

src/mixin.ts
```diff
@@ -73,14 +81,15 @@ export function mixin(Vue: VueConstructor) {
     const setup = vm.$options.setup!
     const ctx = createSetupContext(vm)
 
-    // mark props
-    markReactive(props)
+    // fake reactive for `toRefs(props)`
+    def(props, '__props_reactive__', true)
```
antfu, line 85: Let's put `__props_reactive__` inside `src/utils/symbols.ts` instead of a magic string 😄

antfu (approved):

antfu (approved):

PR #515, opened 2020-09-14
feat(reactive): allow usage of reactive before `Vue.use`

pikax:
Allowing define `reactive` and `ref` before the `Vue.use`, the only caveat is when you have different `vue` versions installed. 

I think this is a pretty edge case and most of the time the user will use the version from `required('vue')`

> Note: would be nice to validate if reactive was used and if it was before the `Vue.use` and the `Vue.use` resolves to a different Vue constructor we could warn it

```js
// something like this
export function setVueConstructor(Vue: VueConstructor) {
  if (__DEV__ && vueConstructor) {
    warn('Another instance of vue installed')
  }

  if(__DEV__ && !vueConstructor && hasReactiveBeenUsed && vueConstructor !== vueDependency) {
    warn('reactive was used globally with a different Vue instance, this might cause reactivity to fail')
  }
  vueConstructor = Vue
  Object.defineProperty(Vue, PluginInstalledFlag, {
    configurable: true,
    writable: true,
    value: true,
  })
}
```

### DIFF
--- a/src/reactivity/reactive.ts
+++ b/src/reactivity/reactive.ts
@@ -1,5 +1,5 @@
 import { AnyObject } from '../types/basic'
-import { getVueConstructor } from '../runtimeContext'
+import { getRegisteredVueOrDefault } from '../runtimeContext'
 import { isPlainObject, def, warn } from '../utils'
 import { isComponentInstance, defineComponentInstance } from '../utils/helper'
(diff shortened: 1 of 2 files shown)


antfu (approved):
This is definitely good to have. Not sure if there are any caveats for bundles though, guess we can just revert it if something really goes wrong. LGTM 👍

PR #526, opened 2020-09-17
fix tsx props autocomplete

xialvjun:


pikax:
Thank you, can you provide an example on what this PR fixes?

xialvjun:
Sorry, this PR seems wrong. But the problem did exist.

```tsx
const SomeCom = defineComponent<{a: number}>({
  props: ['a'],
  setup: (props, ctx) => {
    // props.a is a number;
    return () => <div>{props.a}</div>
  }
});

const App = defineComponent({
  setup: (props, ctx) => {
    return () => (
      // ts doesn't know SomeCom has a prop names a and its type is number
      <SomeCom ></SomeCom>
    )
  }
})
```

but if we write it like:
```tsx
const SomeCom = defineComponent({
  props: ['a'],
  setup: (props, ctx) => {
    // props.a is a number;
    return () => <div>{props.a}</div>
  }
});
```

ts will know SomeCom has a prop names `a` and its type is `any`...

DirtyWhite:
same issue here

Issue #454, opened 2020-08-01
reactive inside computed starts tracking root / top level objects


vmihailenco:
For example see https://codepen.io/vmihailenco/pen/RwrXLGN

Because `data` is not reactive `bar` computed should never be updated. But it looks like call to `reactive` starts tracking changes in `data` (not the whole data - only the top level keys) and because of that `bar` is re-computed every time `data` changes.

Same code works fine in Vue 3 - message is properly toggled on click. `bar` is computed once.

pikax:
This probably is caused because the `reactive` mutates the original object.

After you call `reactive(data)` it will convert `data` into a reactive object.

This is one of the limitations of using `Vue.observe`

![image](https://user-images.githubusercontent.com/4620458/89329301-88f5e700-d686-11ea-8e05-7555900f0ac5.png)

vmihailenco:
@pikax it does not seem to be related to reactive limitation. The same problem applies to computed+ref as well - https://codepen.io/vmihailenco/pen/gOrbNmP . Vue-next works fine.

pikax:
> The same problem applies to computed+ref as well - https://codepen.io/vmihailenco/pen/gOrbNmP

I fail to understand what are you trying to do there.

vmihailenco:
At least it is not reactive limitation this time :)

pikax:
**It is** a limitation of [reactivity in Vue 2.x](https://vuejs.org/v2/guide/reactivity.html#For-Objects)

vmihailenco:
The link you posted is about Vue not able to `detect property addition or deletion`. How it is relevant here?

pikax:
In your example you are adding a new property to the object data and expecting the render to update

antfu:
@vmihailenco `reactive` mutates the object in Vue 2, which means the `var data` you declared becomes reactive after the first run with `computed`, it also collected as a reative dependency of `bar`. So every time you change the `bar.test`, the computed is triggered. Hope this answers your question.

antfu:
Not very sure what you are trying to do here, but if you only want to make the computed run once, try with `const data = markRaw({ test: true });`

vmihailenco:
>In your example you are adding a new property to the object data and expecting the render to update

I do not. See https://codepen.io/vmihailenco/pen/zYqxgYV as even simpler example. And If you check the output you will see that in fact it re-renders on click. The problem is that also unnecessary updates computed. 

[...]

PR #529, opened 2020-09-17
Don't define properties on observed object

vmihailenco:
Potentially fixes https://github.com/vuejs/composition-api/issues/454

1. Are there any problems with this version? Tests seem to pass.
2. Why the same approach can't be used with `reactive`?


pikax:
What are you trying to fix?

PR #532, opened 2020-09-18
fix(reactive): fix issue when using reactive `array` in the template

### LINKED ISSUE #524: Bug in composition-api beta 14
**composition-api beta 12**
```
<template>
  <div>
        <p v-for="item in items">
            {{ item }}
        </p>
  </div>
</template>

export default {
    setup(props, { root }) {
        const items = reactive([]);
        setTimeout(() => {
            items.push(...['a', 'b', 'c']);
        }, 5000);
    }
}
```
**Result:**
```
a
b
c

```
**composition-api beta 14**
`nothing`

pikax:
fix #524

Readded the recommendation of using `ref([])` instead of `reactive([])`, because `reactive([])` doesn't work well with `computed` atm


README.md
```diff
@@ -347,11 +360,9 @@ declare module '@vue/composition-api' {
 </summary>
 
 ```js
-watch(() => {
-    /* ... */
-}, {
+watch(() => {}, {
   immediate: true,
-  onTrack() {}, // not available
+  onTrack() {},
```
antfu, line 365: wow, 👍

README.md
```diff
@@ -110,16 +109,39 @@ export default {
...
+    reactive({
+      count: ref(0),
+    }),
+  ],
+})
+a.list[0].count === 0
+
+a.list.push(
+  reactive({
+    count: ref(1),
```
antfu, line 137: Is the `ref` here really required? A bit too restricted imo.

README.md
```diff
@@ -347,11 +360,9 @@ declare module '@vue/composition-api' {
 </summary>
 
 ```js
-watch(() => {
-    /* ... */
-}, {
+watch(() => {}, {
```
antfu, line 363: Did we miss one arg from the beginning?

README.md
pikax: actually this is a mistake, probably prettier caused it :/

README.md
pikax: I just rolled back this file, currently there's a few issues when using reactive([]) better just tell to use ref on the readme

README.md
```diff
@@ -152,26 +130,55 @@ const a = {
...
 
+// no unwrap for `count`, `.value` is required
 b.list[0].count.value === 0 // true
 ```
 
 </details>
 
+<details>
+<summary>
+✅ <b>Should</b> always use <code>ref</code> in a <code>reactive</code> when working with <code>Array</code>
```
antfu, line 157: Ah, do you mean "If you need to use ref in Array, wrap it with `reactive`"?

pikax: This is the message we had before :D 
https://github.com/vuejs/composition-api/tree/556292d0d60220722fac2cd401b1ba061d6b05ba#ref-unwrap

src/mixin.ts
```diff
@@ -140,6 +148,41 @@ export function mixin(Vue: VueConstructor) {
...
+      defineReactive(target, k, val)
+      if (val) {
+        customReactive(val)
+      }
+      return
+    })
+  }
+
+  function hasReactiveArrayChild(target: object, visited = new Map()): boolean {
+    if (visited.has(target)) return visited.get(target)
```
antfu, line 173:
```suggestion
    if (visited.has(target)) {
      return visited.get(target)
    }
```

src/mixin.ts
```diff
@@ -140,6 +148,43 @@ export function mixin(Vue: VueConstructor) {
...
+    if (visited.has(target)) {
+      return visited.get(target)
+    }
+    visited.set(target, false)
+    if (Array.isArray(target) && isReactive(target)) {
+      visited.set(target, true)
+      return true
+    }
+
+    if (!isPlainObject(target) || isRaw(target)) return false
```
antfu, line 182:
```suggestion
    if (!isPlainObject(target) || isRaw(target)) {
      return false
    }
```

test/setup.spec.js
```diff
@@ -886,4 +886,107 @@ describe('setup', () => {
...
+    const Constructor = Vue.extend(opts).extend({})
+
+    const vm = new Vue(Constructor).$mount()
+    expect(vm.$el.textContent).toBe('[]')
+    await sleep(10)
+    await nextTick()
+    expect(JSON.parse(vm.$el.textContent)).toStrictEqual([{ value: 1 }])
+  })
+
+  // TODO make this pass
```
antfu, line 967: Planned to support it in this PR or another?

README.md
antfu: Not a native English speaker, which statement makes more sense to you? We can improve other statements along the way.

antfu (approved):

PR #538, opened 2020-09-23
fix: Support multiple `composition-api` in one `VueConstructor`

fxxjdedd:
I think that the mechanism that does not allow duplicate registration should be for the same project. 
But there are actually such situations like: two projects share the same `VueConstructor` but use two different composition-api.

This pr solves this problem by adding a random value to the flag.

### DIFF
--- a/src/runtimeContext.ts
+++ b/src/runtimeContext.ts
@@ -13,7 +13,8 @@ try {
 let vueConstructor: VueConstructor | null = null
 let currentInstance: ComponentInstance | null = null
 
-const PluginInstalledFlag = '__composition_api_installed__'
+const randomStr = Math.random().toString(36).substring(2, 15) // Support multiple `composition-api` in one `VueConstructor`
+const PluginInstalledFlag = `__composition_api_${randomStr}_installed__`
 
 export function isPluginInstalled() {
   return !!vueConstructor


pikax:
Why would you want two composition-api libraries to run on the same Vue constructor?

I can't think a reasonable use case

fxxjdedd:
@pikax Here is an example of workspaces:
```
node_modules
    vue
packages
    project1
        node_modules
            composition-api@1.0.0-beta.11
    project2
        node_modules
            composition-api@1.0.0-beta.14
```

The only work-around at present is to unify the versions of these two projects, so that workspaces can place the dependencies to the outer `node_modules`


```
node_modules
    vue
    composition-api@1.0.0-beta.14
packages
    project1
        node_modules
    project2
        node_modules
```

It is always difficult to unify the version. For example, project1 is a git submodule maintained by others

fxxjdedd:
https://github.com/vuejs/composition-api/pull/538#issuecomment-697193988
Sorry, this example is wrong, i will rewrite later

fxxjdedd:
If a lib that set vue as externals uses composition-api, then people using this lib cannot use composition-api any more.

pikax:
The composition-api plugin should be treated as `vue` dependency, it should be only one.

> If a lib that set vue as externals uses composition-api, then people using this lib cannot use composition-api any more.

The lib should always set `vue` and `vue/composition-api` as external, if that's done those dependencies should be correctly resolved by you bundler.

fxxjdedd:
> The lib should always set vue and vue/composition-api as external, if that's done those dependencies should be correctly resolved by you bundler.

This doesn't work. Even as externals, because Vue.use has been executed in the lib, an error will be thrown when Vue.use is executed again in the user code.

However, it is difficult for users to know whether Vue.use has been executed in the lib used.

pikax:
> This doesn't work. Even as externals, because Vue.use has been executed in the lib, an error will be thrown when Vue.use is executed again in the user code.

This has been working on a few libraries - [vue-composable](https://github.com/pikax/vue-composable), [vueuse](https://github.com/antfu/vueuse), etc.

> However, it is difficult for users to know whether Vue.use has been executed in the lib used.

Check how vue-demi solves that: https://github.com/antfu/vue-demi/blob/master/lib/v2/index.esm.js

fxxjdedd:
> The composition-api plugin should be treated as vue dependency, it should be only one.


The premise is that there is only one composition-api as externals in the entire project. However, in actual situations, it is likely that there are multiple composition-apis in the project due to different versions of composition-api. But Vue always has only one.

```
node_modules
    vue
    composition-api@version1
    lib
        composition-api@version2
```

This pr mainly solves this problem.

----

But if the version is unified, the problem is more serious
```
node_modules
    vue
    composition-api
    lib
```


In this case, if we use Vue.use in the lib and user code at the same time, because there is only one composition-api and only one flag, then even this pr cannot be solved.

pikax:
libs should never bundle the composition-api.

fxxjdedd:
> > However, it is difficult for users to know whether Vue.use has been executed in the lib used.

> Check how vue-demi solves that: https://github.com/antfu/vue-demi/blob/master/lib/v2/index.esm.js

Thanks, this is very helpful. But this requires the entire project to share the same composition-api. If there are different versions of composition-api, problems will occur.

fxxjdedd:
> libs should never bundle the composition-api.

I known, I never bundle the composition-api

pikax:
> But this requires the entire project to share the same composition-api

I still don't know what advantages of having multiple composition-api versions, this may only cause a lot of unexpected behaviour.

fxxjdedd:
> I still don't know what advantages of having multiple composition-api versions, this may only cause a lot of unexpected behaviour.

All this is to solve the existing problems of multiple versions, not to benefit from them

pikax:
From what I can understand, this PR is to allow multiple versions to be installed on the same VueContructor, which is in fact having multiple composition-api versions.

You haven't provided a use case where this is required, on the example you provided, I already explained: 

> The composition-api plugin should be treated as `vue` dependency, it should be only one.



> libs should never bundle the composition-api.



If any of the lib you are using is bundling composition-api, please open an issue to their repo.



To end this, having multiple composition-api plugins installed in the same VueContructor, will likely cause some duplication code to run and might cause some unexpected behaviour.


I will close this PR, if you still think this is needed, please write a comment explaining why this needed and what problems it prevents or solves. You can also bring this to [#composition-api discord](https://vue.land)

PR #539, opened 2020-09-23
Fix/global state import order

phbender:


antfu:
Thanks for your contribution. However, we have already clarified the plugin should be installed before the usages. IMHO, there is no need to re-explain it and people may have different ways to organize their code.

https://github.com/vuejs/composition-api/blob/master/README.md#L23

PR #547, opened 2020-10-01
fix: reactive for props

### LINKED ISSUE #534: Bug props in composition-api beta 14
```
export default {
  props: {
    filter: {
      type: Object,
      default: function() {
        return {
          c1: 'c1',
        }
      },
    },
  },
  setup(props) {
    console.log('props is: ', isRef(props), isReactive(props))
    watch(props, () => {  // error 
      console.log(props)
    }
  }
}
```
beta11:  props is: false true
beta14:  props is: false false //  watch --> error

antfu:
fix #534


src/mixin.ts
```diff
@@ -83,7 +82,7 @@ export function mixin(Vue: VueConstructor) {
     const ctx = createSetupContext(vm)
 
     // fake reactive for `toRefs(props)`
-    def(props, PropsReactive, true)
+    def(props, '__ob__', {})
```
antfu, line 85: Or should it be `def(props, '__ob__', (observe({}) as any).__ob__)`?

pikax: using the observable should be "safer", might cause some unexpected side effects, but both would cause that, I would say go for observable for now and see how it goes

src/reactivity/reactive.ts
```diff
@@ -119,6 +119,10 @@ function observe<T>(obj: T): T {
   return observed
 }
 
+export function createObserver() {
+  return (observe({}) as any).__ob__ || {}
```
pikax, line 123:
```suggestion
  return observe<any>({}).__ob__ || {}
```

pikax (approved):

PR #571, opened 2020-10-18
feat(reactivity): add Vue.delete workaround

kiroushi:
Workaround for notifying updates when removing keys in reactive objects and triggering view updates. Chinese README is missing.


antfu:
Looking good. Could you add some unit tests for it? Thanks.

kiroushi:
@antfu there you go.

antfu (approved):

PR #572, opened 2020-10-19
fix: better `vueDependency` importing, close #564

### LINKED ISSUE #564: error Cannot read property 'silent' of undefined
I'm having a problem with it.

My project is in esm mode, so I use @vue/compostition-api esm version of the code.

I define the following code.

```js
import {ref} from '@vue/compostition-api';
const a = ref(false);

export  function useA () {
   return {a};
}; 
```
```js
...
setup(){
  const {a} = useA()
}
```
I got an error.
image
image

When I debugger.
https://github.com/vuejs/composition-api/blob/3a1d99263c91db45fdf90cbd6bb1da9ed2ff20f1/src/runtimeContext.ts#L8
https://github.com/vuejs/composition-api/blob/3a1d99263c91db45fdf90cbd6bb1da9ed2ff20f1/src/runtimeContext.ts#L39
`vueConstructor` is null 
`vueDependency` is {default: ...}

Whether `vueDependency`  should be dealt with. or use `import`, let babel or webpack handle it.

Or there is a problem with my project configuration.  
How it needs to be configured.

antfu:
close #564

### DIFF
--- a/src/runtimeContext.ts
+++ b/src/runtimeContext.ts
@@ -10,10 +10,10 @@ try {
     vueDependency = requiredVue
   } else if (
     requiredVue &&
-    'deafult' in requiredVue &&
-    isVue(requiredVue.deafult)
+    'default' in requiredVue &&
+    isVue(requiredVue.default)
   ) {
-    vueDependency = requiredVue.deafult
+    vueDependency = requiredVue.default
   }
 } catch {
   // not available
--- a/src/runtimeContext.ts
+++ b/src/runtimeContext.ts
@@ -5,7 +5,16 @@ import { assert, hasOwn, warn } from './utils'
 let vueDependency: VueConstructor | undefined = undefined
 
 try {
-  vueDependency = require('vue')
+  const requiredVue = require('vue')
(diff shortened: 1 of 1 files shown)


src/runtimeContext.ts
```diff
@@ -5,7 +5,16 @@ import { assert, hasOwn, warn } from './utils'
...
-  vueDependency = require('vue')
+  const requiredVue = require('vue')
+  if (requiredVue && isVue(requiredVue)) {
+    vueDependency = requiredVue
+  } else if (
+    requiredVue &&
+    'deafult' in requiredVue &&
+    isVue(requiredVue.deafult)
+  ) {
+    vueDependency = requiredVue.deafult
```
kiroushi, line 16:
```suggestion
    'default' in requiredVue &&
    isVue(requiredVue.default)
  ) {
    vueDependency = requiredVue.default
```

Guess this is a typo.

antfu: I must be too sleepy 😓

pikax (approved):

PR #593, opened 2020-11-22
feat: add type-level `readonly()` api

antfu:


src/reactivity/readonly.ts
```diff
@@ -0,0 +1,101 @@
...
+// only unwrap nested ref
+type UnwrapNestedRefs<T> = T extends Ref ? T : UnwrapRef<T>
+
+/**
+ * **In @vue/composition-api, `reactive` only provides type-level readonly check**
+ *
+ * Creates a readonly copy of the original object. Note the returned copy is not
+ * made reactive, but `readonly` can be called on an already reactive object.
+ */
+export function readonly<T extends object>(
```
pikax, line 43: Not sure if is worth to have `readonly` if it's the same as `shallowReadonly`

antfu: It's for the API alignment with Vue 3, lowing the difference between them. Thinking `import { readonly } from 'vue-demi` will work for Vue 3 but failed on Vue 2. This implementation acts like a noop but providing the type-level checking as the same as Vue 3. Should be good enough for users using TypeScript.

src/reactivity/readonly.ts
```diff
@@ -0,0 +1,101 @@
...
+  target: T
+): DeepReadonly<UnwrapNestedRefs<T>> {
+  if (isRef(target)) {
+    return target as any
+  }
+  return shallowReadonly(target) as any
```
antfu, line 49:
```suggestion
  return target as any
```

pikax (approved):

PR #595, opened 2020-11-22
fix(types): improve SetupContext types

antfu:


jacekkarczmarczyk:
shouldn't `slots`, `listeners` and `refs` be using `[keyin string]?: type`?

With `[key: string]` for example `slots.foo` is always defined, but imho user should actually check if there's a slot with provided name

https://www.typescriptlang.org/play?#code/JYOwLgpgTgZghgYwgAgGIHt3IN4ChnIDaA1hAJ4BcyAzmFKAOYC6Vt9IDuAvrrmGQAcUAIThRkAXhz4ipMslA06jJgH5Wyjt14wAriARhg6EMki1kAChiYqGdABpkAIzFVRUAJTSCCE9XQAGwgAOkD0BmtMEJhgAC84510EsIgOMAALJ1coGPjE5LjU9IzPbSA

image

antfu:
Agree. Fixed, thanks!

pikax (approved):

PR #599, opened 2020-11-24
feat: add short hand function: supply, infuse

agrozyme:
I add two short hand function.

**supploy**
for provide multi-set refs (also can for return to template)

```ts
setup() {
  const a = ref('a');
  const b = ref('b');
  const state = reactive({ a, b });
  provide('a', a);
  provide('b', b);
  return { ...toRefs(state) };
}
```

equals

```ts
setup() {
  const a = ref('a');
  const b = ref('b');
  const state = reactive({ a, b });
  return { ...supply(state) };
}
```

**infuse**
for inject multi-set refs

```ts
setup() {
  const state = reactive({ a1: inject('a'), b1: inject('b') });
}
```

equals

```ts
setup() {
  const state = infuse({ a1: 'a', b1: 'b' });
}
```

### DIFF
--- a/src/apis/inject.ts
+++ b/src/apis/inject.ts
@@ -82,7 +82,8 @@ export function supply<T extends Data = Data>(
   const data = <ToRefs<T>>{}
 
   const setValue = (item: any, key: any) => {
-    const value = toRef(item, key)
+    const raw = item[key]
+    const value = isFunction(raw) ? raw : toRef(item, key)
     provide(key, value)
     set(data, key, inject(key))
   }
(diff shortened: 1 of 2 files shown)


antfu:
@agrozyme Thanks for your contribution. But unfortunately, this plugin is trying to mirror Vue 3's composition API, and would not accept additional features that are not presented in Vue 3. If you believe it's a good addtion to Vue's composition API, you might want to follow the [RFC process](https://github.com/vuejs/rfcs) and get the feature landing on Vue 3 first and then porting them back to this plugin. Thanks for understanding.

Issue #604, opened 2020-12-02
IE10 not supporting WeakMap


digitorum:
It using in file `composition-api/src/utils/sets.ts`

antfu:
This plugin requires minimal browser support at IE11 as it's used to move towards Vue 3. If you really need to support lower versions, you may consider polyfill it with your bundler. Thanks.

PR #607, opened 2020-12-03
docs: update readme about Browser Compatibility

digitorum:
https://github.com/vuejs/composition-api/issues/604


antfu:
Can you move the section after ##SSR and before ##Limitations? Thanks!

PR #611, opened 2020-12-07
docs: SetupContext.refs type, remove unnecessary docs

subdavis:
Fix docs bug introduced by https://github.com/vuejs/composition-api/pull/595

## Error

```
ERROR in /home/user/github.com/VIAME-Web/client/platform/web-girder/views/ViewerLoader.vue(16,14):
16:14 Subsequent property declarations must have the same type.  Property 'refs' must be of type 'Data', but here has type '{ [key: string]: Element | Element[] | Vue | Vue[]; }'.
    14 | declare module '@vue/composition-api' {
    15 |   interface SetupContext {
  > 16 |     readonly refs: { [key: string]: Vue | Element | Vue[] | Element[] }
       |              ^
    17 |   }
    18 | }
    19 | /**

```

Type of `refs` changed to unexported type `Data`.  I've verified that this change works on a local project.

### DIFF
--- a/README.md
+++ b/README.md
@@ -375,7 +375,7 @@ import Vue from 'vue'
 
 declare module '@vue/composition-api' {
   interface SetupContext {
-    readonly refs: { [key: string]: Vue | Element | Vue[] | Element[] }
+    readonly refs: { [key: string]: unknown };
   }
 }
 ```
--- a/README.md
+++ b/README.md
@@ -368,18 +368,6 @@ export default {
 }
 ```
 
-You may also need to augment the `SetupContext` when working with TypeScript:
-
-```ts
-import Vue from 'vue'
-
-declare module '@vue/composition-api' {
-  interface SetupContext {
-    readonly refs: { [key: string]: unknown };
-  }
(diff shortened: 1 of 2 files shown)


pikax (changes requested):
`readonly refs: { [key: string]: Vue | Element | Vue[] | Element[] }` is the correct type for refs.

subdavis:
I agree, that _should_ be the correct type.  But it won't build since a conflicting type was added.  Your snippet will not build as a result of #595 via `"^1.0.0-beta.18"`

Either the type needs to change or the docs need to change.  It should probably be corrected here instead:

https://github.com/antfu/composition-api/blob/b7e178e23d75f59cf2af7881debe53911cb57979/src/component/componentOptions.ts#L28

pikax:
Can you change https://github.com/vuejs/composition-api/blob/master/src/component/componentOptions.ts#L28 instead to have the correct type?

pikax (approved):
Thank you

antfu (approved):

PR #627, opened 2021-01-08
fix: Date infer string in props

Zcating:


pikax (approved):
Can you also add it on vue-next? 

https://github.com/vuejs/vue-next/blob/master/packages/runtime-core/src/componentProps.ts#L92-L100

https://github.com/vuejs/vue-next/blob/master/test-dts/defineComponent.test-d.tsx#L36


Thank you

Zcating:
see this issue: [Setting "Date" type as props will infer to string in the setup function](https://github.com/vuejs/vue-next/issues/2668), I copied from it.

Issue #498, opened 2020-09-02
async watchEffect infinite loop with vue 2, fine with vue 3


kabalage:
[Vue 3 example](https://codepen.io/mechast/pen/wvGrgKO?editors=0011)

[Vue 2 example with composition-api](https://codepen.io/mechast/pen/PoNJWzX?editors=0011)

Uncommenting `data.loading = false;` causes infinite loop. I would expect `data.loading` to not be added to the tracked dependencies of the watch effect function.

TeroBlaZe:
This also leads to an infinite loop.
```js
  setup() {
    const state = reactive({
      data: null,
      watchVar: 123,
    });

    const runFunction = (av) => {
        console.log('var', av);
        state.data = []; // <-- this line of code
    }

    watchEffect(() => {
      console.log('watchEffect');
      runFunction(state.watchVar);
    })
  },
```

subdavis:
Could this be elevated in priority?  There are clear reproduction steps.  Thanks!

I've taken to doing this a lot, but it's certainly not ideal:

``` js
// TODO convert to watchEffect
const func = () => {
  // change state
}
watch(value, func);
func();
```

chearon:
It's because of `defineAccessControl`:

https://github.com/vuejs/composition-api/blob/1b6318662df4faddb9cd340b7442f63d013252d6/src/reactivity/reactive.ts#L97

That right there is a Vue getter being called for a mutation of the same property, which establishes a dependency on `a.x` if you do `a.x = 1`. That means, with this bug, `watchEffect` is basically unusable.

Looks like anything set on a `reactive()` object will have this problem, but maybe `Vue.observable()` would be okay if you want a work-around, and don't need the unwrapping. I didn't test that yet, but those are my findings after struggling with this for half a day 😔

PR #638, opened 2021-01-26
test: add failed test for #498

antfu:


milad-mirkarimi (approved):

test/v3/runtime-core/apiWatch.spec.ts
```diff
@@ -531,4 +531,30 @@ describe('api: watch', () => {
...
+      // accessing
+      state.watchVar
+      // setting
+      state.data = []
+      dummy += 1
+    })
+
+    expect(dummy).toBe(1)
+
+    state.data = []
```
chearon, line 554: This shouldn't run the watcher since `state.data` was accessed, not set. A regular Vue `$watch` works that way.

After `watchEffect` I think it should be:

```javascript
expect(dummy).toBe(1)
await nextTick()
expect(dummy).toBe(1)
```

which will fail on `master` and passes on #786

PR #644, opened 2021-02-10
feat: add defineAsyncComponent API

mathe42:
This PR adds a implementation for defineAsyncComponent (I used code from the vue3 implementation) and returns a function async component as used in vue2 (see docs link). This should be a quite similar implementation as vue3 uses and should be compatible in most cases.

Some notes:
1. Tests are a copy of the vue3 tests an then modified.
2. ~Not shure of the typing of `Component` used `VueProxy<any,any>` but maybe there are better options~
3. Changed option suspensible  as vue2 has no suspense component
3. Helpfull docs:
* https://vuejs.org/v2/guide/components-dynamic-async.html#Handling-Loading-State
* https://v3.vuejs.org/api/global-api.html#defineasynccomponent

### DIFF
--- a/src/component/defineAsyncComponent.ts
+++ b/src/component/defineAsyncComponent.ts
@@ -1,18 +1,32 @@
 import { isFunction, isObject, warn } from '../utils'
 import { VueProxy } from './componentProxy'
+import { AsyncComponent } from 'vue'
 
-type Component = VueProxy<any, any>
+import {
+  ComponentOptionsWithoutProps,
+  ComponentOptionsWithArrayProps,
+  ComponentOptionsWithProps,
+} from './componentOptions'
 
-export type AsyncComponentResolveResult<T = Component> = T | { default: T } // es modules
+type ComponentOptions =
+  | ComponentOptionsWithoutProps
+  | ComponentOptionsWithArrayProps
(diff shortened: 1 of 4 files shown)


src/component/defineAsyncComponent.ts
```diff
@@ -0,0 +1,126 @@
...
+          return comp
+        }))
+    )
+  }
+
+  return () => {
+    const component = load()
+
+    return {
+      component: component as any, // there is a type missmatch between vue2 type and the docs
```
mathe42, line 119: Types of Vue2 are not correct in that place I created a PR for that https://github.com/vuejs/vue/pull/11906

antfu (changes requested):
Can you add some runtime tests as well? Thanks.

mathe42:
Will work in it.

Should be easy with the test from vue3 https://github.com/vuejs/vue-next/blob/master/packages/runtime-core/__tests__/apiAsyncComponent.spec.ts ...

mathe42:
Added tests.

Following things are not like vue3:

1. On timeout a Error is displayed.
2. Error component access to thrown Error is not the same.
3. Resolve after timeout doesn't show the component.

mathe42:
Thanks👍

PR #655, opened 2021-03-05
fix(types): RequiredKeys type

### LINKED ISSUE #654: Incorrect type determining required props
I believe we have a bug with how we're determining required vs optional props...

https://github.com/vuejs/composition-api/blob/be9de86a248c8f5e8f52aec1beeee78cc2e3519a/src/component/componentProps.ts#L29-L31

The second half of this condition looks backwards. We should be flagging props as required when they extend `{ default: never }`, not `{ default: any }`. Having a default value means the prop _isn't_ required, correct?

To demonstrate the problem, create a component with the following prop...

```ts
foo: {
  type: String,
  default: 'asdf',
}
```

Then attempt to render that component without defining `foo`.

```
Type '{}' is not assignable to type '{ foo: string; } & {}'.
Property 'foo' is missing in type '{}' but required in type '{ foo: string; }'.
```

As you can see, our optional prop is incorrectly ending up on the required side of [this intersection type](https://github.com/vuejs/composition-api/blob/be9de86a248c8f5e8f52aec1beeee78cc2e3519a/src/component/componentProps.ts#L65-L66).

It looks like this bug also exists in 3.x. If this is correct and I'm not misunderstanding how `default` should be effecting prop types, let me know and I'll open a PR there as well!
https://github.com/vuejs/vue-next/blob/4a965802e883107a2af00301a59fb7f403b6acf7/packages/runtime-core/src/componentProps.ts#L70

scottbedard:
Closes #654

### DIFF
--- a/src/component/componentProps.ts
+++ b/src/component/componentProps.ts
@@ -27,7 +27,9 @@ type PropConstructor<T> =
   | { new (...args: string[]): Function }
 
 type RequiredKeys<T> = {
-  [K in keyof T]: T[K] extends { required: true } | { default: any } ? K : never
+  [K in keyof T]: T[K] extends { required: true } | { default: never }
+    ? K
+    : never
 }[keyof T]
 
 type OptionalKeys<T> = Exclude<keyof T, RequiredKeys<T>>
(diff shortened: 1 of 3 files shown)


antfu (changes requested):
Can you make the tests pass? Thanks!

antfu (approved):

PR #677, opened 2021-04-16
feat(exports): Add ComponentInternalInstance interface to exports

### LINKED ISSUE #675: Missing Export of ComponentInternalInstance
At the moment ComponentInternalInstance is only exported by vue3 not by the vue2 composition api. It is still in the typings. Also getCurrentInstance returns this type but it's not exported by the composition api.

I wanted to ask if has been done for a reason and if not if it could be added?

relief-melone:
- Added ComponentInternalInstance to exports
- Version bump to rc7
- Updated rollup-plugin-typescript2 to v0.30.0
- Added yarn to dev dependencies
- Added tslib to dev dependencies

Fixes #675

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@vue/composition-api",
-  "version": "1.0.0-rc.6",
+  "version": "1.0.0-rc.7",
   "description": "Provide logic composition capabilities for Vue.",
   "keywords": [
     "vue",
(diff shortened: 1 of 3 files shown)


antfu:
Thank you but please only do one thing at a time. We use the script to bump the version there is no need to do that manually. Thanks for the contributions but sorry I can't merge it. Will fix it in another commit.

relief-melone:
I can split the pull requests. No problem. Should be no problem now that I know about the versioning

Issue #636, opened 2021-01-23
[TypeScript] boolean optional prop is typed as `foo?: boolean`


jacekkarczmarczyk:
Following code

```ts
// main.ts
import Vue from 'vue';
import { defineComponent } from '@vue/composition-api';

let bool = false;

Vue.extend({
  props: { foo: Boolean },
  mounted () {
    bool = this.foo; // No error
  },
});

defineComponent({
  props: { foo: Boolean },
  setup (props) {
    bool = props.foo; // <!--- Error
  },
});
```

throws a TS error:

```
  Type 'undefined' is not assignable to type 'boolean'.
    18 |   props: { foo: Boolean },
    19 |   setup (props) {
  > 20 |     bool = props.foo;
       |     ^
    21 |   },
    22 | });
```

danielroe:
This looks like the correct behaviour to me.

jacekkarczmarczyk:
> This looks like the correct behaviour to me.

Doesn't look for me, if prop is not passed its value is bool: https://codepen.io/jkarczm/pen/VwKOEzp

pikax:
Is not the correct behaviour, if `Boolean` is passed, it will be implicit converted to `false` if not provided

jacekkarczmarczyk:
that can be closed i guess?

PR #678, opened 2021-04-17
fix(types): optional Boolean prop types #636

almost-no-one:
#636


antfu (approved):
Matched with latest vue-next

PR #680, opened 2021-04-18
feat(typing): Add ComponentInternalInstance interface to exports

### LINKED ISSUE #675: Missing Export of ComponentInternalInstance

relief-melone:
ComponentInternalInstance added to exports

Resolves #675

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -6,7 +6,7 @@ export const version = __VERSION__
 
 export * from './apis'
 export * from './component'
-export { getCurrentInstance } from './runtimeContext'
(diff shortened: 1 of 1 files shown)


antfu:
It's already been fixed and released, Thank you.

https://github.com/vuejs/composition-api/blob/12b544e544af4c1b4118ac7c872ecff49626020d/src/index.ts#L9

PR #684, opened 2021-04-23
fix: Memory leak caused by global variables.

ygj6:
Fix:[#685](https://github.com/vuejs/composition-api/issues/685)


leealso:
Was this ever fixed?

Issue #685, opened 2021-04-23
vue + compositionApi will cause memory leaks!


ygj6:
### Version
2.6.12

### Reproduction link
[https://codesandbox.io/s/billowing-butterfly-qe2yv?file=/src/App.vue](https://codesandbox.io/s/billowing-butterfly-qe2yv?file=/src/App.vue)

### Steps to reproduce
See [codesandbox](https://codesandbox.io/s/billowing-butterfly-qe2yv?file=/src/App.vue) for details, When the `close` button is clicked, the Vuecomponent memory will not be released compared to before the click.

![Memory leaks](https://user-images.githubusercontent.com/7699524/115829655-d21ede80-a441-11eb-921f-12e48ebfe053.PNG)

I found that when the `setup() {},` of the child component is commented out, the memory can be released normally.
### What is expected?
Could release memory normally.

### What is actually happening?
Memory leaks!

ygj6:
@pikax @antfu 
The detailed reproduction steps are:
1. Download the demo in the [link](https://codesandbox.io/s/billowing-butterfly-qe2yv?file=/src/App.vue) and execute `npm install ` and `npm run serve`;
2. Open the local URL ;
3. Open the developer tools, click the `take heap snapshort` button under the memory tab, and generate a memory snapshot `A `
4. Click the `close` button on the page to destroy the component, and repeat step 3 to generate a memory snapshot `B`;

 Then we will find: `B` does not release memory compared to `A`


![Memory leaks](https://user-images.githubusercontent.com/7699524/116495327-44c20b00-a8d5-11eb-887b-14b7ea66c99a.PNG)

I think this is a memory leak bug, can you take a moment to look at this issue and PR?

PR #686, opened 2021-04-23
fix: Memory leak caused by global variables.

ygj6:
Fix: #685

### DIFF
--- a/src/utils/utils.ts
+++ b/src/utils/utils.ts
@@ -14,21 +14,17 @@ export const hasSymbol =
 
 export const noopFn: any = (_: any) => _
 
-const sharedPropertyDefinition = {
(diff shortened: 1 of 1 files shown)


ygj6:
https://github.com/vuejs/composition-api/blob/12b544e544af4c1b4118ac7c872ecff49626020d/src/mixin.ts#L250-L256
https://github.com/vuejs/composition-api/blob/12b544e544af4c1b4118ac7c872ecff49626020d/src/utils/utils.ts#L17-L32

When mounting the `setter` and `getter`, a reference to the current vue instance by the global variable `sharedPropertyDefinition` appears.

antfu (approved):

Issue #669, opened 2021-03-25
`shallowReadonly` does not work for `ref`s


vvanpo:
## Description

`shallowReadonly()` successfully prevents mutations on objects created with `reactive()`, but not `ref()`.

## Minimal reproduction

Vue 2 example (broken):
https://codesandbox.io/s/example-forked-ofgpu?file=/src/App.vue

Constrast this with the equivalent implentation in Vue 3, which works as expected:
https://codesandbox.io/s/example-forked-olke4?file=/src/App.vue

Note how in the `@vue/composition-api` example, assigning to the `countRefReadonly` value does not report an error and mutates the underlying `ref`. Whereas in Vue 3, the `ref` is not mutated and the assignment triggers a warning.

PR #704, opened 2021-05-24
fix(shallowReadonly) : shallowReadonly should work for ref.

ygj6:
fix: #669


test/v3/reactivity/readonly.spec.ts
```diff
@@ -372,5 +374,32 @@ describe('reactivity/readonly', () => {
...
+        setup() {
+          const count = reactive({ number: 0 })
+          const countRef = ref(0)
+
+          const countReadonly = shallowReadonly(count)
+          const countRefReadonly = shallowReadonly(countRef)
+
+          setTimeout(() => {
+            // @ts-ignore
+            countReadonly.number++
```
iheyunfei, line 391: I would suggest `@ts-expected-error` would be a better option.

ygj6: Thanks!

PR #706, opened 2021-05-26
fix(types): should unwrap tuple correctly

iheyunfei:
same problem in vuejs/vue-next#3819


antfu:
Maybe it's better to be fixed in vue-next first https://github.com/vuejs/vue-next/blob/90aa8358129f25826bfc4c234325c1442aef8d55/packages/reactivity/src/ref.ts#L222-L222

iheyunfei:
@antfu Yes, I sent a PR to fix it vuejs/vue-next#3820 .

PR #722, opened 2021-06-04
feat(reactivity): unwrap value when using `set`

### LINKED ISSUE #721: Ref not unwrapping other refs set with `set` on Node
- Clone https://github.com/posva/bug-report-vca
- `yarn`
- `node index.js`

It should log an object `{ old: false, new: true }` but it logs `{ old: false, new: { value: true }}`

There is also an index.html file to show the (expected) behavior on a browser

pikax:
fix #721

When using `set` if the object is not `reactive` it won't execute the unwrapping as per expected on a reactive object, in vue 2 every object used is expected to be `reactive` by default so I believe is safe to assume that if you calling `set` we will want to unwrap the value if is a `ref`

### DIFF
--- a/src/reactivity/set.ts
+++ b/src/reactivity/set.ts
@@ -34,7 +34,8 @@ export function set<T>(target: any, key: any, val: T): T {
     return val
   }
   if (!ob) {
-    target[key] = val
+    // If we are using `set` we can assume that the unwrapping is intended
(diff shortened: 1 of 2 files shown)


posva (approved):
Nice!

antfu (approved):

PR #723, opened 2021-06-05
fix(observe): solve the Ref not unwrapping on the ssr side issue with recursive way.

ygj6:
Solving #721 with #722：
```
  it('result of #722', () => {
    const state = { old: ref(false) }
    set(state, 'new', ref(true))
    expect(JSON.stringify(state)).toBe(
      '{"old":{"value":false},"new":true}'  // 'old' is different from 'new'
    )
  })
```
the form of value of `old` key is different from `new` key.

if we solve #721 with recursive way:
```
  it('solving with recursive way', () => {
    const state = { old: ref(false) }
    set(state, 'new', ref(true))
    expect(JSON.stringify(state)).toBe(
      '{"old":{"value":false},"new":{"value":true}}'
    )
  })
```
for above case, should `{"old":{"value":false},"new":{"value":true}}` is more appropriate? @pikax @antfu Look forward your comments？

### DIFF
--- a/src/reactivity/reactive.ts
+++ b/src/reactivity/reactive.ts
@@ -118,9 +118,28 @@ export function observe<T>(obj: T): T {
   // in SSR, there is no __ob__. Mock for reactivity check
   if (!hasOwn(observed, '__ob__')) {
     def(observed, '__ob__', mockObserver(observed))
+    walk(observed)
   }
 
   return observed
+}
+
+/**
+ * Recursive obj, add __ob__
+ */
+function walk(obj: any) {
+  var keys = Object.keys(obj)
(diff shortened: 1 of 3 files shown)


src/reactivity/reactive.ts
```diff
@@ -118,11 +118,30 @@ export function observe<T>(obj: T): T {
...
+    walk(observed)
   }
 
   return observed
 }
 
+/**
+ * Recursive obj, add __ob__
+ */
+function walk(obj: any) {
```
antfu, line 130:
```suggestion
function walk(obj: any, seen = new WeakMap<any, boolean>()) {
```

Would need a map to avoid circular reference.

PR #724, opened 2021-06-05
fix: only allow watch callbacks to be self-triggering

### LINKED ISSUE #633: watchEffect is not same as vue3 watchEffect?
```
setup() {
    const num = ref(0);
    watchEffect(() => {
      console.log("num change will not trigger this effect");
      num.value = 1;
    });
    setTimeout(() => {
      num.value++;
    }, 5000);
    return {
      num
    };
  }
```
in vue3, it's right, watchEffect only triggered by value getter(not setter), in composition-api, it will triggered by value setter which means it's not right

vue3-demo https://stackblitz.com/edit/vue-qfisdz
sorry, I did't create a demo for @vue/composition-api.

ygj6:
fix #633

### DIFF
--- a/src/apis/watch.ts
+++ b/src/apis/watch.ts
@@ -1,3 +1,4 @@
+import Vue from 'vue'
 import { ComponentInstance } from '../component'
 import { Ref, isRef, isReactive } from '../reactivity'
 import {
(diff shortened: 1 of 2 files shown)


src/apis/watch.ts
```diff
@@ -248,6 +250,21 @@ function createWatcher(
   // effect watch
   if (cb === null) {
     let running = false
+    const { Dep } = getVueInternalClasses()
+    Dep.prototype.notify = function notify() {
```
pikax, line 254: We should avoid patching `prototype` methods.

 This patching of the prototype will run every time we call a watch, because of the scoped `running` it might cause problems.

pikax (changes requested):

PR #726, opened 2021-06-07
chore: warning message grammar

Flamenco:
Use present participle ('providing' instead of 'provide')

### DIFF
--- a/src/reactivity/reactive.ts
+++ b/src/reactivity/reactive.ts
@@ -143,7 +143,7 @@ export function shallowReactive<T extends object = any>(obj: T): T
 export function shallowReactive(obj: any): any {
   if (!isObject(obj)) {
     if (__DEV__) {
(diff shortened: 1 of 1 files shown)


src/reactivity/reactive.ts
```diff
@@ -143,7 +143,7 @@ export function shallowReactive<T extends object = any>(obj: T): T
 export function shallowReactive(obj: any): any {
   if (!isObject(obj)) {
     if (__DEV__) {
-      warn('"shallowReactive()" is called without provide an "object".')
+      warn('"shallowReactive()" is called without providing an "object".')
```
posva, line 146: I think we could use this opportunity to improve the whole wording:

```suggestion
      warn('"shallowReactive()" must be called on an object.')
```

@antfu @pikax WDYT?

pikax: This sounds good to me

Issue #418, opened 2020-06-30
Watched Previous Values can't be Destructure on First Fire


klondikemarlen:
When? watching an array of values I can't destructure the previous values on the initial loop run.
Why? the previous values are returned as `undefined`.
What I expect? I expect the previousValue on initial run to be equal to `[undefined, undefined]` so I can destructure it properly.

Code I have to write:
```javascript
watch(
  () => [unref(unitId), unref(entityType)],
  ([newUnitId, newEntityType], oldValues) => {
    const [oldUnitId, oldEntityType] = oldValues || [undefined, undefined]

    if (!newUnitId || !newEntityType) return
    if (
      entitySchemaState.entitySchema &&
      newUnitId === oldUnitId &&
      newEntityType === oldEntityType
    )
      return

    fetch()
  },
  { immediate: true }
)
```

Code I want to write:
```javascript
watch(
  () => [unref(unitId), unref(entityType)],
  ([newUnitId, newEntityType], [oldUnitId, oldEntityType]) => {
    if (!newUnitId || !newEntityType) return
    if (
      entitySchemaState.entitySchema &&
      newUnitId === oldUnitId &&
      newEntityType === oldEntityType
    )
      return

    fetch()
  },
  { immediate: true }
)
```

pikax:
That's the same behaviour as you have in vue3.

AFAIK this is the expected behaviour since you are deconstructing an `undefined`.

antfu:
@klondikemarlen This should works for you

```ts
watch(
  [unitId, entityType],
  ([newUnitId, newEntityType], [oldUnitId, oldEntityType]) => {

  }
)
```

antfu:
Hope it solves your problem. Closing for now.

klondikemarlen:
I know that I can't destructing an undefined. :P

I want the watch function to _always_ return an array if it is operating on an array of args.

klondikemarlen:
> @klondikemarlen This should work for you
> 
> ```ts
> watch(
>   [unitId, entityType],
>   ([newUnitId, newEntityType], [oldUnitId, oldEntityType]) => {
> 
>   }
> )
> ```

When I try passing in an array instead of a function, I get the same destructuring error.
![image](https://user-images.githubusercontent.com/23045206/86516200-6b0d3c00-bdd3-11ea-8758-bb2066d60802.png)

antfu:
I will have a look

MrDenexi:
[...]

PR #727, opened 2021-06-08
fix(watch): watched previous values can't be destructure on first fire.

ygj6:
Fix: #418, align behavior with vue-next.

### DIFF
--- a/src/apis/watch.ts
+++ b/src/apis/watch.ts
@@ -329,6 +329,7 @@ function createWatcher(
     // The subsequent callbacks will redirect to `callback`.
     let shiftCallback = (n: any, o: any) => {
       shiftCallback = originalCallback
+      o = isArray(n) ? [] : undefined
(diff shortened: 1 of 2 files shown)


src/apis/watch.ts
```diff
@@ -329,6 +329,7 @@ function createWatcher(
     // The subsequent callbacks will redirect to `callback`.
     let shiftCallback = (n: any, o: any) => {
       shiftCallback = originalCallback
+      o = isArray(n) ? [] : undefined
       applyCb(n, o)
```
posva, line 333: I think it's better not to modify the argument
```suggestion
      // o is undefined on the first call
      applyCb(n, isArray(n) ? [] : o)
```

posva (changes requested):
Nice!

posva (approved):

pikax (approved):
Thank you

PR #731, opened 2021-06-15
chore: upgrade vue to 2.6.14

ygj6:
This PR and https://github.com/vuejs/vue/pull/12132 work together to solve the problem of compilation warnings after the `vue` in `composition-api` is upgraded to the latest version.

Before https://github.com/vuejs/vue/pull/12132 is merged, we will see that ci fails to run.

### DIFF
--- a/src/env.d.ts
+++ b/src/env.d.ts
@@ -15,18 +15,4 @@ declare module 'vue/types/vue' {
     _watchers: VueWatcher[]
     __composition_api_state__?: VfaState
   }
-
-  interface VueConstructor {
-    observable<T>(x: any): T
-    util: {
-      warn(msg: string, vm?: Vue | null)
-      defineReactive(
-        obj: Object,
-        key: string,
-        val: any,
-        customSetter?: Function,
-        shallow?: boolean
-      )
-    }
-  }
 }
--- a/src/env.d.ts
+++ b/src/env.d.ts
@@ -15,4 +15,18 @@ declare module 'vue/types/vue' {
     _watchers: VueWatcher[]
     __composition_api_state__?: VfaState
   }
+
+  interface VueConstructor {
+    observable<T>(x: any): T
+    util: {
+      warn(msg: string, vm?: Vue | null)
+      defineReactive(
+        obj: Object,
+        key: string,
+        val: any,
+        customSetter?: Function,
+        shallow?: boolean
+      )
+    }
+  }
 }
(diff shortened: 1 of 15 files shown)


src/env.d.ts
```diff
@@ -15,18 +15,4 @@ declare module 'vue/types/vue' {
     _watchers: VueWatcher[]
     __composition_api_state__?: VfaState
   }
-
-  interface VueConstructor {
-    observable<T>(x: any): T
-    util: {
-      warn(msg: string, vm?: Vue | null)
-      defineReactive(
```
posva, line 23: Isn't there a way to keep this code without changing the public api on vuejs/vue code?

ygj6: @posva Maybe we could use `// @ts-ignore` or use `any` to define the type to avoid the warnings.

ygj6:
> This PR and [vuejs/vue#12132](https://github.com/vuejs/vue/pull/12132) work together to solve the problem of compilation warnings after the `vue` in `composition-api` is upgraded to the latest version.
> 
> Before [vuejs/vue#12132](https://github.com/vuejs/vue/pull/12132) is merged, we will see that ci fails to run.

@antfu  @posva Now it is solved by shielding the warning, no longer dependent on [vuejs/vue#12132](https://github.com/vuejs/vue/pull/12132)

posva:
After further investigation, your initial solution (https://github.com/vuejs/vue/pull/12132) is probably a better idea. Merging that one would allow to not change anything here, right?

ygj6:
> After further investigation, your initial solution ([vuejs/vue#12132](https://github.com/vuejs/vue/pull/12132)) is probably a better idea. Merging that one would allow to not change anything here, right?

Here I have deleted invalid module extension of `VueConstructor`:
https://github.com/vuejs/composition-api/pull/731/files#:~:text=interface%20VueConstructor%20%7B,defineReactive(

Because they are already defined in `vue/types/vue`, so the extension here will not take effect.

I noticed that you added the `null` type here：
https://github.com/vuejs/composition-api/commit/dd2cd6b56e6a6cc454391b9e0d07291a5bbcc623#:~:text=export%20function%20warn(msg%3A%20string%2C%20vm%3F%3A%20Vue%20%7C%20null)%20%7B

but there is no `null` type in `Vue`: https://github.com/vuejs/vue/blob/656a3cbda089c26938802707223254990631b91f/types/vue.d.ts#L125

 which will cause a warning, so I temporarily deleted the `null` here: https://github.com/vuejs/composition-api/pull/731/files#:~:text=export%20function%20warn(msg%3A%20string%2C%20vm%3F%3A%20Vue%20%7C%20null)

If don't want to delete `null`, we also could add `null` type in `Vue`. @posva Which solution do you think is better?

PR #732, opened 2021-06-16
fix(reactivity): check type of __ob__ in isRaw and isReactive

### LINKED ISSUE #715: Incorrect check isReactive for Proxy objects
I have this code that using some DI container:
```
setup() {
   const readApi = container.resolve(ReadApi);

   return { readApi }
}
```

In my tests I replace value, provided by container, with mock object, created using jest-mock-extended. Under the box it's just a `Proxy` object.

It works well until I updated to version `rc.9`.
Here we have this change: https://github.com/vuejs/composition-api/commit/8c27d80#diff-40aa26982716adc75031925664c28d5fb4fc6dac8c7ce4d6f5122b61c1ed773eR27
And now I have such errors: `Cannot read property 'depend' of undefined`.
That's because isReactive checks for `__ob__` property of object, which is exists, but it's a mock function. So, `propValue.__ob__.dep` is undefined. Maybe you can add additional check in isReactive or just do something like `isReactive(propValue) && propValue.__ob__.dep)`

Djaler:
fixes #715 

We should check the type of `__ob__` because in some cases it may be not an expected object, but, for example, a mock function, so the fact that `Boolean(!__ob__.__raw__) === true` does not necessarily indicate that the object is reactive

### DIFF
--- a/src/reactivity/reactive.ts
+++ b/src/reactivity/reactive.ts
@@ -15,11 +15,15 @@ import { isRef, UnwrapRef } from './ref'
 import { rawSet, accessModifiedSet } from '../utils/sets'
 
 export function isRaw(obj: any): boolean {
-  return Boolean(obj?.__ob__ && obj.__ob__?.__raw__)
+  return Boolean(
+    obj?.__ob__ && typeof obj.__ob__ === 'object' && obj.__ob__?.__raw__
+  )
 }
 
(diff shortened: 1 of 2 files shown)


pikax (approved):
Thank you

Djaler:
BTW, this problem is not present in Vue 3, so you can treat this PR as behavior alignment with vue-next)

PR #755, opened 2021-07-08
fix: only trigger warning in the dev environment

webfansplz:
fix: only trigger warning in the dev environment


src/utils/helper.ts
```diff
@@ -33,7 +33,7 @@ export function isComponentInstance(obj: any) {
 
 export function createSlotProxy(vm: ComponentInstance, slotName: string) {
   return (...args: any) => {
-    if (!vm.$scopedSlots[slotName]) {
+    if (__DEV__ && !vm.$scopedSlots[slotName]) {
```
antfu, line 36: This changes the behaviour, should return silently on prod

webfansplz: fixed

PR #765, opened 2021-07-16
fix(readonly): align behavior with vue-next.

ygj6:


src/reactivity/readonly.ts
```diff
@@ -44,6 +44,11 @@ type UnwrapNestedRefs<T> = T extends Ref ? T : UnwrapRef<T>
 export function readonly<T extends object>(
   target: T
 ): DeepReadonly<UnwrapNestedRefs<T>> {
+  if (!isObject(target)) {
+    if (__DEV__) {
```
antfu, line 48: This API is provided for type level only, it's better to judge `__DEV__` first

ygj6: sure

PR #798, opened 2021-08-19
bug(watch): don't leak memory in Options API hooks

### LINKED ISSUE #797: Watchers created inside Options API hooks should be bound to the current Vue 2 instance
In Vue 3, you can call `watch`, `watchEffect`, etc. from Options API hooks like `created`, and they will be torn down when the VM is destroyed. It's highly useful to use these functions outside of just the `setup` API because they can be imported and used by supporting libraries.

In `@vue/composition-api`, you can do this, but the VM that gets used is the fallback VM, which is global, so doing this is effectively a memory leak.

I have a test and a patch that I'll submit soon.

chearon:
Tracks Vue 2 active instances by adding to the start and end of the hook list in `beforeCreate`, if there are any hooks there (if there are any hooks check is for performance). Adds a check in watch functions for a Vue 2 instance if there isn't one in effect scope or one for `setup()`.

I could also re-use the existing `activeInstance` mechanism, but that creates a whole Vue 3-like instance so it could slow down components with hooks.

Fixes #797

### DIFF
--- a/src/apis/watch.ts
+++ b/src/apis/watch.ts
@@ -13,7 +13,7 @@ import {
   isMap,
 } from '../utils'
 import { defineComponentInstance } from '../utils/helper'
-import { getVueConstructor } from '../runtimeContext'
+import { getVueConstructor, getCurrentVue2Instance } from '../runtimeContext'
 import {
   WatcherPreFlushQueueKey,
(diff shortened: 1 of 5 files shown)


chearon:
See discussion in #797

PR #820, opened 2021-10-01
feat(computed): allow differentiating refs from computed

posva:
Add an `effect` property on computed to differentiate them from refs

---

Another possible implementation is to keep the computed types where they were and add an option effect to ref


posva:
I don't understand why the node v12 is failing 😓

PR #795, opened 2021-08-19
test: add failing test of isReactive

posva:
I only added the failing test because I don't know where this should be fixed. I imagine something related to `defineAccessControl()` but I couldn't figure it out yet so feel free to to add commits to the PRs or create a new one

Note the same code works in client and on Vue 3

PR #821, opened 2021-10-01
fix(ssr): `set()` twice lose reactivity

posva:
Same as #795 

It seems to still be failing.

It creates https://github.com/posva/pinia/issues/648

### DIFF
--- a/test/ssr/ssrReactive.spec.ts
+++ b/test/ssr/ssrReactive.spec.ts
@@ -59,10 +59,22 @@ describe('SSR Reactive', () => {
 
   it('should work on objects sets with set()', () => {
     const state = ref<any>({})
-    set(state.value, 'a', {})
-    set(state.value, 'a', {})
 
+    set(state.value, 'a', {})
     expect(isReactive(state.value.a)).toBe(true)
+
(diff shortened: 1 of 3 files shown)


posva:
So it works 🤔 
I think it has to deal with the rename of the main branch

What is weird is that the error "toRefs() expects a reactive object" appears on the server on Nuxt even though it's running 1.2.2


Regarding the test, should we add this? It doesn't seem like it was added back then

test/ssr/ssrReactive.spec.ts
```diff
@@ -57,6 +57,13 @@ describe('SSR Reactive', () => {
     expect(isRaw(state)).toBe(false)
   })
 
+  it('should work on objects sets with set()', () => {
+    const state = ref<any>({})
+    set(state.value, 'a', {})
```
posva, line 62:
```suggestion
    set(state.value, 'a', {})
    set(state.value, 'a', {})
```

This seems to make a difference

posva:
Found the problem: setting twice. I'm going to fix this on Pinia but I don't think this edge case is expected.

PR #834, opened 2021-10-16
Return a more reasonable type for arrays

cheese-git:
Actually it caused a type error in my command line, which complained for missing properties:  [Symbol.iterator], [Symbol.unscopables].


cheese-git:
Thanks reply, I just opened a PR in vue-next 😄 

https://github.com/vuejs/vue-next/pull/4807

xiaoxiangmoe:
@cheese-git This PR will break tuple type

xiaoxiangmoe:
Close it now. Reopen it when vue 3 support this.

Issue #850, opened 2021-11-15
Can't get instance type from component definition


Djaler:
I have this utility function for cases when I know for sure that template ref is exists. Also, it provides me correct typing for this ref methods.
```
type VueClass<V extends Vue = Vue> = (new (...args: any[]) => V) & typeof Vue

export function useComponentRef<T extends VueClass>() {
    return ref<InstanceType<T>>(undefined!) as Ref<InstanceType<T>>;
}
```
Usage:
```
const foo = useComponentRef<typeof Foo>();
 
onMounted(() => {
    foo.value.bar(); // method bar strictly typed 
}
 ```

But after update to 1.4.0 this doesn't work.
Now I receive this error:
```
TS2344: Type 'VueProxy<{}, void, Data, {}, {}, {}, {}, {}>' does not satisfy the constraint '(new (...args: any[]) => Vue) & VueConstructor<Vue>'.   Type 'VueProxy<{}, void, Data, {}, {}, {}, {}, {}>' is not assignable to type 'new (...args: any[]) => Vue'.     Construct signature return types 'ComponentRenderProxy<{} & {}, void, Data, {}, {}, {}, {}, {}, {} & {}, {}, true>' and 'Vue' are incompatible.       The types of '$attrs' are incompatible between these types.         Type 'Data' is not assignable to type 'Record<string, string>'.
```
I found that it was broken because of this change: https://github.com/vuejs/composition-api/commit/b980175426a25400a5d0ba9f684eee44901fae65#diff-5d5edc98cf929d645f9b70ae45d64d3a58088caddcc1b99f433b256aa93d4a90R91
If I replace `Omit<VueConstructor, never>` back to `VueConstructor`, my typing works good.

MinatoHikari:
Yea, because of that change, type of VueProxy does not satisfy the constraint VueConstructor, I can't mixins Components defined bt `defineComponent` with `vue-class-component`

xiaoxiangmoe:
@Djaler If we change `Omit<VueConstructor, never>` to `VueConstructor`, volar type check for $emits will fail.

xiaoxiangmoe:
@Djaler If we change `Omit<VueConstructor, never>` to `VueConstructor`, volar type checking for $emits will fail.

xiaoxiangmoe:
You can rewrite it as follows:
```ts
export function useComponentRef<T extends new () => T>() {
[...]

PR #869, opened 2021-12-06
fix: resolve conflicts with vue2 interface

MinatoHikari:
https://github.com/vuejs/composition-api/issues/850

Vue2 renderComponent Interface is different from vue3 now.

It works for me to resolve type error when I use Mixins in vue-class-component

Maybe we can change it's type again after vue2.7

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,6 +38,7 @@
     "lint": "prettier --write --parser typescript \"{src,test,test-dts}/**/*.ts?(x)\" && prettier --write \"{src,test}/**/*.js\"",
     "test": "yarn test-dts && yarn test-unit",
     "test-unit": "cross-env NODE_ENV=test jest",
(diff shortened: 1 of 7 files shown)


xiaoxiangmoe (approved):
LGTM

xiaoxiangmoe:
@MinatoHikari Can you add some type test for this in directory `/test-dts`? Thank you very much.

MinatoHikari:
One problem, type checking in vsCode or webStorm is fine but vue-tsc will be failed when using vue3 tsconfig.
@antfu @xiaoxiangmoe

xiaoxiangmoe:
@MinatoHikari Maybe we can remove vue-class-component and vue-property-decorator in devDependencies. Only extract simpler use case from it.

MinatoHikari:
Remove vue-class-component and vue-property-decorator.
To avoid conflicts when using vue3 type,  move some tests to `defineComponent-vue2.d.tsx`, add exclude file for `tsconfig.vue3.json`.

xiaoxiangmoe (approved):

Issue #874, opened 2021-12-20
[BUG] setup doesn't support return object with type of Module 


xialvjun:
The bug( or maybe not bug) code is at https://github.com/vuejs/composition-api/blob/a43090d098b37388798d6db12cc2292177828b49/src/mixin.ts#L118

Sorry I can not offer an codesandbox or stackblitz link, because this bug is related to the building enviroment.

My local build enviroment is: vite + vite-plugin-vue2;

The build script is just `"build": "vite build",` without `tsc &&` because esbuild support typescript.

Then with code:

App.vue
```vue
<template><div @click="a++">{{a}}</div></template>
<script lang="ts">
import { defineComponent } from "@vue/composition-api"
import * as refs from "./refs"; // refs is just a ts file with some global ref state.

export default defineComponent({
  setup() {
    return { ...refs }; // refs is of type "[object Module]", and `{ ...refs }` is also of type "[object Module]", so isPlainObject return false.
    // I need to convert it to
    const spread = obj => Object.keys(obj).reduce((acc, cv) => ({...acc, [cv]: obj[cv]}), {});
    return { ...(spread(refs)) }; // then it works.
  }
});
</script>
```

refs.ts
```ts
import {} from '@vue/composition-api';
export const a = ref(0);
export const b = ref('');
```

MinatoHikari:
You should not directly export const something = ref()
```

export const useRef = () => {
    const a = ref()

    return { a }
}
```

LinusBorg:
@MinatoHikari exporting a ref like OP does is totally fine.

xialvjun:
I don't have a need to SSR, so I just export a global ref.

xialvjun:
I think changing `isPlainObject` to `val => val !== null && typeof val === 'object'` is enough, because vue3 is just like this https://github.com/vuejs/vue-next/blob/87c86e4cc2/packages/runtime-core/src/component.ts#L688

PR #875, opened 2021-12-23
fix: #874 setup return object with type of Module

xialvjun:
`isPlainObject` is too strict for the returning value of setup function. Sometimes user may `return { ...refs }` in setup function, and `Symbol(Symbol.toStringTag) in refs` (eg: refs is a ESModule object), then `{ ...refs }` will inherit that symbol property, then `isPlainObject({ ...refs }) === false` which may not that intuitive.

Besides, vue3 use `isObject` too. https://github.com/vuejs/vue-next/blob/87c86e4cc2/packages/runtime-core/src/component.ts#L688

### DIFF
--- a/src/mixin.ts
+++ b/src/mixin.ts
@@ -115,7 +115,7 @@ export function mixin(Vue: VueConstructor) {
         return activateCurrentInstance(instance, () => bindingFunc())
       }
       return
-    } else if (isPlainObject(binding)) {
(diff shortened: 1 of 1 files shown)


xialvjun:
@antfu If you thought one word change is not worth a PR, will you tell me where to fix this problem? Or will you fix it in next release? I issued this PR just to push things change. It seems no one is talking in #874

PR #894, opened 2022-01-25
fix(shallowReactive): don't trigger watchers for oldVal === newVal

chearon:
Hi, I found another situation that's causing me infinite loops with this library but not with Vue 3

### DIFF
--- a/src/reactivity/ref.ts
+++ b/src/reactivity/ref.ts
@@ -184,7 +184,9 @@ export function shallowRef(raw?: unknown) {
 export function triggerRef(value: any) {
   if (!isRef(value)) return
 
-  value.value = value.value
+  const v = value.value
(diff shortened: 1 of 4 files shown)


src/reactivity/ref.ts
```diff
@@ -184,7 +184,9 @@ export function shallowRef(raw?: unknown) {
 export function triggerRef(value: any) {
   if (!isRef(value)) return
 
-  value.value = value.value
+  const v = value.value
+  value.value = !v
+  value.value = v
```
chearon, line 189: might need some advice here... this is not ideal for `sync: true` watchers, but I can't call `ob.notify()` without making bigger changes as the reference is not retrievable from `value`

antfu (approved):

PR #986, opened 2023-08-09
feat: support inject key types

hawtim:
add inject key types, correct me if I got wrong.


hawtim:
tests passed, no types error.

image


image

image

antfu (approved):

